PR #1, opened 2020-03-02
gRPC implementation for Mino - Handler.Process

nkcr:
First step that only implements the simple Handler.Process

This is far from perfect and there are still untested edge cases, but that would be good to do an early review.


go.mod
```diff
@@ -3,9 +3,17 @@ module go.dedis.ch/fabric
 go 1.13
 
 require (
+	cloud.google.com/go v0.53.0
+	github.com/desertbit/timer v0.0.0-20180107155436-c41aec40b27f // indirect
+	github.com/gogo/protobuf v1.3.1
```
Gilthoniel: Can you check where it is used and replace it by the correct one ?

go.mod
```diff
@@ -3,9 +3,17 @@ module go.dedis.ch/fabric
 go 1.13
 
 require (
+	cloud.google.com/go v0.53.0
```
Gilthoniel: what is that ?

mino/minogrpc/mod.go
```diff
@@ -1,3 +1,86 @@
...
+
+// MakeNamespace creates a new Minogrpc struct that has the specified namespace.
+// This namespace is further used to scope newly created RPCs.
+//
+// TODO: decide if a server can evolve through multiple different namespaces or
+// if a pre-condition of this function is to have an empty namespace.
+func (m Minogrpc) MakeNamespace(namespace string) (mino.Mino, error) {
+	newM := Minogrpc{
+		server:    m.server,
+		namespace: namespace,
```
Gilthoniel: My idea at the beginning was to have ns1/ns2/ns3/.../rpc

mino/minogrpc/mod.go
```diff
@@ -1,3 +1,86 @@
...
+
+	return rpc, nil
+}
+
+// MakeMinoGrpc sets up the grpc and http servers. It does not start the
+// server. Identifier must be an address with a port, something like
+// 127.0.0.1:3333
+//
+// TODO: use a different type of argument for identifier, maybe net/url ?
+func MakeMinoGrpc(identifier string) (*Minogrpc, error) {
```
Gilthoniel: The functions are implemented by value so you don't need to return a pointer. Also can we declare the constructor after the struct and using NewSomething so we keep the same naming ?

mino/minogrpc/server.go
```diff
@@ -0,0 +1,296 @@
...
+}
+
+// StartServer makes the server start listening and wait until its started
+func (srv *Server) StartServer() error {
+
+	go func() {
+		err := srv.Serve()
+		// TODO: better handle this error
+		if err != nil {
+			log.Fatal("failed to start server ", err)
```
Gilthoniel: `fabric.Logger`

mino/minogrpc/server.go
```diff
@@ -0,0 +1,296 @@
...
+				errs <- xerrors.Errorf("failed to get client conn: %v", err)
+				continue
+			}
+			cl := NewOverlayClient(clientConn)
+
+			ctx, ctxCancelFunc := context.WithTimeout(context.Background(),
+				defaultContextTimeout)
+			defer ctxCancelFunc()
+
+			header := metadata.New(map[string]string{"apiuri": rpc.uri})
```
Gilthoniel: Why is that necessary ?

mino/minogrpc/mod.go
nkcr: The annoying thing is that we cannot return `nil, xerrors....` if it's not a pointer

mino/minogrpc/server.go
nkcr: It is used to pass the handler's URI to the gRPC call.

mino/minogrpc/overlay.go
```diff
@@ -0,0 +1,61 @@
...
+// Call is the implementation of the overlay.Call proto definition
+func (o overlayService) Call(ctx context.Context, msg *CallMsg) (*CallResp, error) {
+	// We fetch the uri that identifies the handler in the handlers map with the
+	// grpc metadata api. Using context.Value won't work.
+	headers, ok := metadata.FromIncomingContext(ctx)
+	if !ok {
+		return nil, xerrors.Errorf("header not found in provided context")
+	}
+
+	apiURI, ok := headers["apiuri"]
```
Gilthoniel: Can you add a constant for that string ?

mino/minogrpc/server.go
```diff
@@ -0,0 +1,296 @@
...
+	priv, err := ecdsa.GenerateKey(elliptic.P521(), rand.Reader)
+	if err != nil {
+		return nil, xerrors.Errorf("Couldn't generate the private key: %+v", err)
+	}
+
+	tmpl := &x509.Certificate{
+		SerialNumber: big.NewInt(1),
+		IPAddresses:  []net.IP{net.ParseIP("127.0.0.1")},
+		NotBefore:    time.Now(),
+		NotAfter:     time.Now().Add(time.Hour * 24 * 180),
```
Gilthoniel: Also a constant here.

nkcr: Ok, that's done

PR #4, opened 2020-03-06
Setup Travis + coverage

Gilthoniel:
Setup Travis with coverage to help the early stages.


Gilthoniel:
Apparently we have both Travis(.org|.com) configured for the organization. I disabled on .com so it should trigger only for .org for the next PRs.

PR #7, opened 2020-03-11
Hash module

Gilthoniel:
This adds a hash factory to not rely on a single hash algorithm and improve unit tests further.

### DIFF
--- a/consensus/cosipbft/mod.go
+++ b/consensus/cosipbft/mod.go
@@ -32,12 +32,17 @@ type Consensus struct {
 
 // NewCoSiPBFT returns a new instance.
 func NewCoSiPBFT(mino mino.Mino, cosi cosi.CollectiveSigning) *Consensus {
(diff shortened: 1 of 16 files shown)


consensus/cosipbft/mod.go
```diff
@@ -32,12 +32,17 @@ type Consensus struct {
...
 	c := &Consensus{
 		storage: newInMemoryStorage(),
 		mino:    mino,
 		cosi:    cosi,
-		factory: newChainFactory(cosi.GetVerifier()),
-		queue:   &queue{verifier: cosi.GetVerifier()},
+		factory: chainFactory,
+		queue: &queue{
+			verifier: cosi.GetVerifier(),
+			factory:  chainFactory,
```
nkcr: this is not directly in the scope of this PR, but I would find it much easier to follow (especially since there are a lot of different factory thing) if `factory` would be renamed to `chainFactory`.

PR #8, opened 2020-03-11
Refactor cosi interface

Gilthoniel:
This refactors the cosi interface so that it can verify the signatures during the protocol. The actual unit tests will come after this PR.

### DIFF
--- a/blockchain/skipchain/block.go
+++ b/blockchain/skipchain/block.go
@@ -279,7 +279,10 @@ func (f *blockFactory) decodeConodes(factory crypto.PublicKeyFactory, msgs []*Co
 }
 
 func (f *blockFactory) decodeBlock(factory crypto.PublicKeyFactory, src proto.Message) (SkipBlock, error) {
-	in := src.(*BlockProto)
+	in, ok := src.(*BlockProto)
+	if !ok {
+		return SkipBlock{}, xerrors.Errorf("unknown message type: '%T'", src)
+	}
 
 	var payload ptypes.DynamicAny
 	err := protoenc.UnmarshalAny(in.GetPayload(), &payload)
--- a/blockchain/skipchain/block.go
+++ b/blockchain/skipchain/block.go
@@ -66,6 +66,11 @@ func (b SkipBlock) GetHash() []byte {
 	return b.hash[:]
 }
 
+// GetPreviousHash returns the previous block digest.
+func (b SkipBlock) GetPreviousHash() []byte {
+	return b.BackLinks[0].Bytes()
+}
+
 // GetPublicKeys returns the list of public keys of the block.
 func (b SkipBlock) GetPublicKeys() []crypto.PublicKey {
 	pubkeys := make([]crypto.PublicKey, 0, b.Conodes.Len())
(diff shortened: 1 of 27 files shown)


blockchain/skipchain/member.go
```diff
@@ -48,25 +49,76 @@ func (c Conode) Pack() (proto.Message, error) {
 	return conode, nil
 }
 
+type iterator struct {
+	conodes []Conode
+	index   int
+}
+
+func (i *iterator) Next() bool {
```
nkcr: ~Next~ HasNext should only check `if i.index < len(i.conodes)` and ~Get~ GetNext should return the conode AND increment the index.

consensus/cosipbft/requests.go
```diff
@@ -0,0 +1,102 @@
...
+	req.digest, err = forwardLink.computeHash(f.New())
+	if err != nil {
+		return req, xerrors.Errorf("couldn't compute hash: %v", err)
+	}
+
+	return req, nil
+}
+
+// GetHash returns the hash of the proposal that will be signed by the collective
+// authority.
```
nkcr:
```suggestion
// GetHash returns the hash of the prepare request that will be signed by the
// collective authority.
```

consensus/cosipbft/requests.go
```diff
@@ -0,0 +1,102 @@
...
+	commit := Commit{
+		to:      to,
+		prepare: prepare,
+		hash:    buffer,
+	}
+
+	return commit, nil
+}
+
+// GetHash returns the bytes of the prepare signature.
```
nkcr: this comment is hard to understand: here we are returning the hash of the "request sent for the last phase of the PBFT". So `c.Hash` is actually the bytes of the prepare signature?

consensus/cosipbft/requests.go
```diff
@@ -0,0 +1,102 @@
...
+func (c Commit) GetHash() []byte {
+	return c.hash
+}
+
+// Pack returns the protobuf message or an error,
```
nkcr:
```suggestion
// Pack returns the protobuf message representation of a Commit, or an error
```

cosi/mod.go
```diff
@@ -3,13 +3,22 @@ package cosi
...
+// PublicKeyIterator is an iterator over the list of public keys of a
+// collective authority.
+type PublicKeyIterator interface {
+	Next() bool
+	Get() crypto.PublicKey
+}
+
+// CollectiveAuthority (or Cothority in short) is a set of participant to a
+// collective signature.
+type CollectiveAuthority interface {
```
nkcr: I appreciate the use of a non-abbreviated word 😇

mino/minoch/manager.go
```diff
@@ -3,44 +3,55 @@ package minoch
...
+		return nil
+	}
+
+	return m.instances[string(text)]
 }
 
 func (m *Manager) insert(inst *Minoch) error {
 	addr := inst.GetAddress().(address)
-	if addr.String() == "" {
+	text, err := addr.MarshalText()
```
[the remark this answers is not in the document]
Gilthoniel: That's the standard golang interface for marshaling text.

blockchain/skipchain/block.go
```diff
@@ -66,9 +68,14 @@ func (b SkipBlock) GetHash() []byte {
 	return b.hash[:]
 }
 
-// GetPublicKeys returns the list of public keys of the block.
-func (b SkipBlock) GetPublicKeys() []crypto.PublicKey {
-	return b.Conodes.GetPublicKeys()
+// GetPreviousHash returns the previous block digest.
+func (b SkipBlock) GetPreviousHash() []byte {
+	return b.BackLinks[0].Bytes()
```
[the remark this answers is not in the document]
Gilthoniel: A skipblock can only come from the factory from an external point of view and the factory makes sure that it is defined (among other checks).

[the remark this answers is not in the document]
Gilthoniel: (it will, not quite yet implemented)

PR #9, opened 2020-03-13
Cosi unit tests

Gilthoniel:
This PR adds the unit tests for the former blscosi renamed to flatcosi as it does not depend on the crypto algorithm. It also slightly refactors cosi so that the Listen function returns an actor. The same will be done for the consensus in a later PR.

Also: add a testing helper to cover protobuf messages.


cosi/flatcosi/mod.go
```diff
@@ -0,0 +1,153 @@
...
+				if err != nil {
+					return nil, xerrors.Errorf("couldn't verify the aggregation: %v", err)
+				}
+
+				return agg, nil
+			}
+
+			agg, err = a.processResponse(resp, agg)
+			if err != nil {
+				a.logger.Err(err).Msg("error when processing response")
```
nkcr: not sure why you sometimes return an error or just log it. A small explanation could help.

cosi/mod.go
```diff
@@ -35,12 +28,17 @@ type Message interface {
 	GetHash() []byte
 }
 
+// Actor is the listener of a collective signing instance. It provides a
+// primitive to sign a message.
+type Actor interface {
```
nkcr: note for later: add `Actor` in the list of terminologies. All those terminologies can be confusing at first: Actor, Players, Verifier, ect...

crypto/bls/mod.go
```diff
@@ -14,119 +16,144 @@ import (
...
 //go:generate protoc -I ./ --go_out=./ ./messages.proto
 
-var suite = pairing.NewSuiteBn256()
+var (
+	suite = pairing.NewSuiteBn256()
+
+	protoenc encoding.ProtoMarshaler = encoding.NewProtoEncoder()
+)
 
+// publicKey is the implementation of the crypto.PublicKey interface for a BLS
```
nkcr: For a later refactoring, I suggest we systematically use `<element> implements <abstraction>(). It ...` if its an interface function and `<element> implements <abstraction>{}. It ...` if it's a struct.

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.13
 
 require (
 	github.com/desertbit/timer v0.0.0-20180107155436-c41aec40b27f // indirect
+	github.com/gogo/protobuf v1.3.1
```
nkcr: oops ?

cosi/flatcosi/mod.go
Gilthoniel: Good point. That is because the collective signing should keep going even if one participant is doing weird stuff. But some errors are critical and should stop the protocol.

cosi/mod.go
Gilthoniel: Yes, I'll add it now.

go.mod
Gilthoniel: gosh I hate that module

PR #13, opened 2020-03-17
Improves stream

nkcr:
Adds test case with multiple servers.
The stream is still limited in that clients can only reply to the caller.


mino/minogrpc/server.go
```diff
@@ -135,20 +145,27 @@ func (rpc RPC) Call(req proto.Message,
 	return out, errs
 }
 
-// Stream implements mino.RPC.Stream()
+// Stream implements mino.RPC.Stream.
```
Gilthoniel: I was about to make a PR to change the doc for this. IMHO, it does not make sense to repeat Stream there as it is stated at the beginning and just below. `Stream implements mino.RPC. It ...` is enough to get all the information in one place.

mino/minogrpc/overlay.proto
```diff
@@ -18,4 +18,8 @@ service Overlay {
 
 message OverlayMsg {
     google.protobuf.Any message = 1;
+}
+
+message DummyMsg {
```
Gilthoniel: Could you use a wrapper available? https://godoc.org/github.com/golang/protobuf/ptypes/wrappers
So we don't have a test message publicly available ?

mino/minogrpc/server.go
```diff
@@ -135,20 +145,27 @@ func (rpc RPC) Call(req proto.Message,
...
 func (rpc RPC) Stream(ctx context.Context,
 	players mino.Players) (in mino.Sender, out mino.Receiver) {
 
-	errs := make(chan error, 1)
+	// if every player produces an error the buffer should be large enought so
+	// that we are never blocked in the for loop and we can termninate this
+	// function.
+	errs := make(chan error, players.Len())
```
Gilthoniel: btw that's the reason we can't pass only an iterator ;)

mino/minogrpc/server.go
```diff
@@ -159,7 +176,10 @@ func (rpc RPC) Stream(ctx context.Context,
 		if err != nil {
 			// TODO: try another path (maybe use another node to relay that
 			// message)
-			errs <- xerrors.Errorf("failed to get client conn: %v", err)
+			fabric.Logger.Error().Msgf("failed to get client conn for client '%s': %v",
+				addr.String(), err)
+			errs <- xerrors.Errorf("failed to get client conn for client '%s': %v",
+				addr.String(), err)
```
Gilthoniel:
```go
err := ...
fabric.Logger.Err(err).Send()
errs <- err
```

mino/minogrpc/server.go
```diff
@@ -171,7 +191,10 @@ func (rpc RPC) Stream(ctx context.Context,
 
 		stream, err := cl.Stream(ctx)
 		if err != nil {
-			errs <- xerrors.Errorf("failed to get stream from client: %v", err)
+			fabric.Logger.Error().Msgf("failed to get stream for client '%s': %v",
+				addr.String(), err)
+			errs <- xerrors.Errorf("failed to get stream for client '%s': %v",
+				addr.String(), err)
```
Gilthoniel: same as above

mino/minogrpc/server.go
```diff
@@ -181,13 +204,25 @@ func (rpc RPC) Stream(ctx context.Context,
...
+				msg, err := stream.Recv()
+				if err == io.EOF {
+					return
+				}
+				status, ok := status.FromError(err)
+				if ok && err != nil && status.Code() == codes.Canceled {
+					return
+				}
+				if err != nil {
+					fabric.Logger.Error().Msgf("failed to receive for client '%s': %v",
```
Gilthoniel: same as above

mino/minogrpc/server.go
```diff
@@ -379,20 +403,28 @@ func (s sender) Send(msg proto.Message, addrs ...mino.Address) error {
...
 		err = player.streamClient.Send(sendMsg)
+		if err == io.EOF {
+			return nil
+		}
 		if err != nil {
 			return xerrors.Errorf("failed to call the send on client stream: %v", err)
 		}
 	}
 
+	if !ok {
```
Gilthoniel: In the case of a stream, it's fair to return an error if something bad happened I guess. You could use a go routine per address in the for loop and gather possible error/nil in an array then create an error from the list if any.

mino/minogrpc/server_test.go
```diff
@@ -104,11 +103,10 @@ loop:
 	}
 
 	server.grpcSrv.GracefulStop()
-	err = server.httpSrv.Shutdown(context.Background())
 	require.NoError(t, err)
 }
 
-func Test_ErrorsSimpleCall(t *testing.T) {
+func TestErrorsSimpleCall(t *testing.T) {
```
Gilthoniel: I try to name my tests using this format: TestTypeName_FunctionName. It is easier to find to which piece of code it is related. Here it could be: TestRPC_Call. What do you think, could we add that to the guidelines ?

mino/minogrpc/server_test.go
```diff
@@ -464,19 +450,237 @@ func Test_ErrorsSimpleStream(t *testing.T) {
...
+	}
+
+	// the handler must be registered on each server. Fron the client side, that
+	// means the "registerNamespace" and "makeRPC" must be called on each
+	// server.
+	server1.handlers[uri] = handler
+	server2.handlers[uri] = handler
+	server3.handlers[uri] = handler
+
+	dummyMsg := &DummyMsg{Value: "dummy_value"}
```
Gilthoniel: `dummyMsg := &wrappers.StringValue{Value: "dummy_value"}`

mino/minogrpc/server.go
nkcr: Then we should change the phrasing, because `Stream implements mino.RPC.Stream` describes precisely what it does, while saying `Stream implements mino.RPC` is not exactly right.
To be right it should be something like `Stream implements one of the function described by mino.RPC.Stream`.

mino/minogrpc/overlay.proto
nkcr: cool, I was precisely looking for something like that!

mino/minogrpc/server_test.go
nkcr: yep, makes sense!

mino/minogrpc/server.go
Gilthoniel: That's exactly what `Stream implements mino.RPC.` means. You implement an interface by implementing the functions of the interface. Yes it's one of the functions, but you know which one.

nkcr: > You implement an interface by implementing the functions of the interface

True for a struct, not a function. A struct can say that it implements an interface, but a function not, this is why I am not comfortable saying `Stream implements mino.RPC.`. (and this is why in the first place I was using "()", to indicate that the abuse of saying "blabla implements interface.blabla()" actually means this is *contributing to* implementing  the interface, but does not actually *implements* it, since it's a function.)

Anyway, we can agree to disagree on this one and keep going with your proposed phrasing.

mino/minogrpc/server.go
nkcr: but in this case this *could* be a normal behaviour. I am displaying a warning because this is very likely to be an undesirable situation, but I can imagine situations where it is normal that no senders are included in the list of participants.

Gilthoniel: I'm not sure to get the purpose. It will warn if none of the address matches but if you provide a wrong address, it will silently ignore and you have no way to know one was wrong. You can check the length at the beginning.

mino/minogrpc/server.go
```diff
@@ -339,27 +361,29 @@ func makeCertificate() (*tls.Certificate, error) {
...
+	name         string
 }
 
 type player struct {
 	address      address
 	streamClient overlayStream
 }
 
-// send implements mino.Sender.Send()
+// send implements Mino.Sender.Send
```
Gilthoniel: It's a package, not the Mino interface thus `mino` is correct.

mino/minogrpc/server.go
nkcr: Then I could add another warning at line 383 ?  I did not because excluding someone that is not in the list of address is a normal behavior, while excluding everyone is an orange flag.

Gilthoniel: Thus my point of collecting the errors that are happening in each iteration and then make a summary of the results.

nkcr: Ok. Changing the Send interface to return a `chan error`

PR #15, opened 2020-03-18
Update guideline for comments for interface implementation.

Gilthoniel:


nkcr:
For the record, we went through many possibilities:

```go
// Stream belongs to the mino.RPC implementation
// Stream implements mino.RPC.Stream
// Stream implements mino.RPC.Stream()
// Stream implements mino.RPC
// Stream contributes to mino.RPC
// Stream contributes to implementing mino.RPC
// Stream belongs to the mino.RPC interface
// Stream is an element of the mino.RPC interface
// Stream is a member of mino.RPC
// Stream belongs to mino.RPC
// Stream belongs to interface mino.RPC
// Stream belongs to mino.RPC interface
```

And we choose to favour a concise and easy to remember phrasing with `// Stream implements mino.RPC` even if that may not be a 100% technically true. We believe it is better to have simple rules that are followed instead of complex ones that we don't use. 
Some discussion here https://github.com/dedis/fabric/pull/13#discussion_r394138988.

docs/guideline.md
```diff
@@ -43,6 +43,23 @@ type RPC struct {
 
 ## Tests
 
+Tests should be named according to their specific scope of testing. For
+instance, a unit test of a function should follow that convention:
```
nkcr:
```suggestion
Tests should be named according to their specific scope of testing, following 
the syntax `Test<scope>_<function>`. For instance, a unit test of a function 
should follow that convention:
```

docs/guideline.md
```diff
@@ -43,6 +43,23 @@ type RPC struct {
...
+
+func (d Dummy) String() string {...}
+```
+
+```go
+// dummy_test.go
+
+func TestDummy_String(t *testing.T) {}
+```
+
```
nkcr: I would add something to handle more complex situation, like:

For some situation, it is necessary to add detailed information about a test case. This can be done between the SCOPE and the FUNCTION:

```go
func Test_RPC_MultipleSimple_Stream(t *testing.T) {
...
}
```

As an invariant, the SCOPE and FUNCTION should always be the first, respectively the last, element.

Gilthoniel: Done

PR #16, opened 2020-03-18
Skipchain unit tests

Gilthoniel:
This PR adds unit tests to most of the skipchain package. Remaining files will be done later on.


blockchain/mod.go
```diff
@@ -38,6 +38,10 @@ type Validator interface {
 
 // Actor is a primitive created by the blockchain to propose new blocks.
 type Actor interface {
+	// InitChain creates a new chain with data as the payload and the players as
+	// the first roster.
+	InitChain(data proto.Message, players mino.Players) error
```
nkcr: Is it creating a first block with "data" as the payload of this first block? If so we should clarify it.

blockchain/skipchain/validator.go
```diff
@@ -0,0 +1,79 @@
...
+// - implements consensus.Validator
+type blockValidator struct {
+	*Skipchain
+
+	validator blockchain.Validator
+	buffer    SkipBlock
+}
+
+// Validate implements consensus.Validator. It decodes the message into a block
+// and validate its integrity. It returns the block if it is correct, otherwise
```
nkcr:
```suggestion
// and validates its integrity. It returns the block if it is correct, otherwise
```

blockchain/skipchain/validator.go
```diff
@@ -0,0 +1,79 @@
...
+		return nil, encoding.NewDecodingError("block", err)
+	}
+
+	genesis, err := v.db.Read(0)
+	if err != nil {
+		return nil, xerrors.Errorf("couldn't read genesis block: %v", err)
+	}
+
+	if !bytes.Equal(genesis.GetHash(), block.GenesisID.Bytes()) {
+		return nil, xerrors.Errorf("mismatch genesis hash %v != %v", genesis.hash, block.GenesisID)
```
nkcr: If it happens that the hash is empty, the error message can be misleading:
"mismatch genesis hash  != aef123.."
```suggestion
		return nil, xerrors.Errorf("mismatch genesis hash '%v' != '%v'", genesis.hash, block.GenesisID)
```

blockchain/mod.go
```diff
@@ -38,6 +38,10 @@ type Validator interface {
 
 // Actor is a primitive created by the blockchain to propose new blocks.
 type Actor interface {
+	// InitChain creates a new chain with the given data stored as the payload
+	// and the given players to be the roster of the genesis block.
```
nkcr:
```suggestion
	// InitChain initializes a new chain by creating the genesis block with the
	// given data stored as the payload and the given players to be the roster
	// of the genesis block.
```

PR #17, opened 2020-03-19
Close consensus executions

Gilthoniel:
This PR adds a close function to consensus actors to abort execution.

### DIFF
--- a/consensus/cosipbft/mod.go
+++ b/consensus/cosipbft/mod.go
@@ -74,6 +74,7 @@ func (c *Consensus) Listen(v consensus.Validator) (consensus.Actor, error) {
 	}
 
 	actor := pbftActor{
+		closing:     make(chan struct{}),
 		hashFactory: c.factory.GetHashFactory(),
 	}
 
@@ -92,14 +93,16 @@ func (c *Consensus) Listen(v consensus.Validator) (consensus.Actor, error) {
 }
 
 type pbftActor struct {
+	closing     chan struct{}
 	hashFactory crypto.HashFactory
 	cosiActor   cosi.Actor
 	rpc         mino.RPC
 }
 
-// Propose takes the proposal and send it to the participants of the consensus.
-// It returns nil if the consensus is reached and that the participant are
-// committed to it, otherwise it returns the refusal reason.
+// Propose implements consensus.Actor. It takes the proposal and send it to the
+// participants of the consensus. It returns nil if the consensus is reached and
+// that the participant are committed to it, otherwise it returns the refusal
+// reason.
 func (a pbftActor) Propose(p consensus.Proposal, players mino.Players) error {
 	prepareReq, err := newPrepareRequest(p, a.hashFactory)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


consensus/cosipbft/mod.go
```diff
@@ -92,14 +93,16 @@ func (c *Consensus) Listen(v consensus.Validator) (consensus.Actor, error) {
...
 	rpc         mino.RPC
 }
 
-// Propose takes the proposal and send it to the participants of the consensus.
-// It returns nil if the consensus is reached and that the participant are
-// committed to it, otherwise it returns the refusal reason.
+// Propose implements consensus.Actor. It takes the proposal and send it to the
+// participants of the consensus. It returns nil if the consensus is reached and
+// that the participant are committed to it, otherwise it returns the refusal
+// reason.
```
nkcr:
```suggestion
// Propose implements consensus.Actor. It takes the proposal and sends it to the
// participants of the consensus. It returns nil if the consensus is reached and
// the participant are committed to it, otherwise it returns the refusal reason.
```

consensus/mod.go
```diff
@@ -53,6 +53,8 @@ type Actor interface {
 	// Propose performs the consensus algorithm using the list of nodes
 	// as participants.
 	Propose(proposal Proposal, players mino.Players) error
+
+	Close() error
```
nkcr: Should the close be used to signal a normal ending operation or an unexpected event? In the later case "Abort()" would be more appropriate.

consensus/qsc/mod.go
```diff
@@ -67,38 +69,59 @@ func (c *Consensus) GetChain(id []byte) (consensus.Chain, error) {
...
+				proposal = nil
+			}
+
+			ctx, cancel := context.WithTimeout(context.Background(), EpochTimeout)
+
+			go func() {
+				// This Go routine is responsible for listening a close event
+				// from the actor.
+				select {
+				case <-ctx.Done():
```
nkcr: no check of `ctx.Error()`?

consensus/qsc/mod.go
```diff
@@ -67,38 +69,59 @@ func (c *Consensus) GetChain(id []byte) (consensus.Chain, error) {
...
-		// TODO: how to stop the loop ?
 		for {
-			err := c.executeRound(val)
+			var proposal consensus.Proposal
+			select {
+			case <-c.closing:
+				fabric.Logger.Trace().Msg("closing")
+				return
+			case proposal = <-c.ch:
+			default:
```
nkcr: That seems weird: if there isn't any proposals in `c.ch`, then this loops forever at full speed calling executeRound() with a nil proposal?

consensus/mod.go
Gilthoniel: No it's for normal close.

consensus/qsc/mod.go
Gilthoniel: If there is an error, it will be propagated by the `executeRound` function and then logged.

consensus/qsc/mod.go
Gilthoniel: Well, either you have a proposal or none but you need to participate to the consensus anyway. It odes not loop at full speed because you have a time step execution every time.

Gilthoniel:
@nkcr commented and fixed

PR #19, opened 2020-03-19
minogrpc: Implements message relaying

nkcr:
Adds a simple relaying mechanism: if a client must send a message to someone it doesn't know, it packs the message in an enveloppe and sends it to the orchestrator that relays the message. 

More tests to increase coverage will come.


mino/minogrpc/overlay.proto
```diff
@@ -8,7 +8,7 @@ import "google/protobuf/any.proto";
 message Envelope {
     string from = 1;
     repeated string to = 2;
-    google.protobuf.Any message = 3;
+    google.protobuf.Any message = 4;
```
Gilthoniel: ?

mino/minogrpc/server.go
```diff
@@ -187,7 +187,7 @@ func (rpc RPC) Stream(ctx context.Context,
 
 		header := metadata.New(map[string]string{
 			headerURIKey:     rpc.uri,
-			headerAddressKey: ""})
+			headerAddressKey: addr.String()})
```
Gilthoniel: Can you add a comment to explain what is this address ?

mino/minogrpc/server.go
```diff
@@ -215,21 +215,60 @@ func (rpc RPC) Stream(ctx context.Context,
...
+					}
+				} else {
+					anyEnvelope, err := ptypes.MarshalAny(&Envelope{
+						Message: msg.Message})
+					if err != nil {
+						fillAndLog(errs, encoding.NewAnyEncodingError(anyEnvelope, err))
+						return
+					}
+
+					orchRecv.in <- &OverlayMsg{Message: anyEnvelope}
```
Gilthoniel: Do you really need to have both types of message ? Isn't an envelop simply an overlay message ?

mino/minogrpc/server.go
```diff
@@ -215,21 +215,60 @@ func (rpc RPC) Stream(ctx context.Context,
...
+				}
 			}
 		}()
 	}
 
 	return orchSender, orchRecv
 }
 
+func relayMessage(enveloppe *Envelope, orchSender sender) error {
+	var dynamicAny ptypes.DynamicAny
```
Gilthoniel: It looks like it is not used.

mino/minogrpc/server.go
```diff
@@ -472,17 +549,26 @@ func (r receiver) Recv(ctx context.Context) (mino.Address, proto.Message, error)
 
 	enveloppe := &Envelope{}
 	err = ptypes.UnmarshalAny(msg.Message, enveloppe)
-	if err != nil {
-		return nil, nil, encoding.NewAnyDecodingError(msg.Message, err)
+	// we can receive two types of messages: enveloppes and other messages. If
```
Gilthoniel: Why don't you make the logic easier by always sending an envelope/overlay message ? When a message is received, you check that the address is yours or if you should relay.

mino/minogrpc/server_test.go
```diff
@@ -688,66 +655,171 @@ func TestRPC_MultipleChange_Stream(t *testing.T) {
...
+	dummyMsg2, ok := msg2.(*wrappers.StringValue)
+	require.True(t, ok)
+	require.Equal(t, "dummy_value_"+identifier1+"_"+identifier2+"_"+identifier3+"_"+identifier4,
+		dummyMsg2.Value)
+
+	server1.grpcSrv.GracefulStop()
+	server2.grpcSrv.GracefulStop()
+	server3.grpcSrv.GracefulStop()
+	server4.grpcSrv.GracefulStop()
+	require.NoError(t, err)
```
Gilthoniel: what error ?

mino/minogrpc/server.go
nkcr: line 227 ; )

mino/minogrpc/server.go
nkcr: When a server wants to send a message to someone it doesn't know, it then creates an enveloppe with the "from" and "to" and sends it back. However, a server does not know its own address, this is why we set it in the header metadata.

If I make a better use of the enveloppe messages, I should be able to get rid of this attribute (the sever could just read the "to" attribute of the enveloppe to know its own address).

mino/minogrpc/server.go
nkcr: True, I was getting closer to this conclusion. I am just concerned that using only enveloppes will be insufficient for more advanced routing. What do you think?

mino/minogrpc/server_test.go
nkcr: oops, that's the rest of a non-graceful server stop

mino/minogrpc/server.go
Gilthoniel: I'm not sure to understand why a server couldn't know it's own address. Can't you instantiate the service with the address given when instantiating mino ?

nkcr: that's definitely doable!

PR #24, opened 2020-03-23
RPC: add a context for cancellation

Gilthoniel:
This adds a context to the Call function such that the caller can cancel the request if necessary.

### DIFF
--- a/mino/mod.go
+++ b/mino/mod.go
@@ -27,15 +27,26 @@ type AddressIterator interface {
 // Players is an interface to represent a set of nodes participating in a
 // message passing protocol.
 type Players interface {
+	// Take should a subset of the players according to the filters.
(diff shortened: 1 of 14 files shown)


mino/mod.go
```diff
@@ -27,15 +27,26 @@ type AddressIterator interface {
...
+	// AddressIterator returns an iterator that prevents changes of the
+	// underlying array and save memory by iterating over the same array.
 	AddressIterator() AddressIterator
+
+	// Len returns the length of the set of players.
 	Len() int
 }
 
+// ChanMessages is an alias for a channel of Protobuf messages.
+type ChanMessages = <-chan proto.Message
```
nkcr: I would definitely not make aliases in this case.

PR #25, opened 2020-03-23
Skipchain leader only

Gilthoniel:
This adds information to the Handler.Process so that network layer info are accessible. It then uses that information to force a skipchain block to be proposed by the leader only.


mino/mod.go
```diff
@@ -38,6 +39,14 @@ type Sender interface {
 	Send(msg proto.Message, addrs ...Address) <-chan error
 }
 
+// Request is a wrapper around the context of a message received from a player
+// and that needs to be processed by the node. It provides some useful
+// information about the network layer.
+type Request struct {
+	Address Address
```
nkcr: I would comment what address this corresponds to

mino/mod.go
```diff
@@ -59,7 +68,7 @@ type RPC interface {
 type Handler interface {
 	// Process handles a single request by producing the response according to
 	// the request message.
-	Process(req proto.Message) (resp proto.Message, err error)
+	Process(req Request) (resp proto.Message, err error)
```
nkcr: following that, shouldn't we also update Receiver.Recv from `Recv(context.Context) (Address, proto.Message, error)` to `Recv(context.Context) (Request, error)` ?

nkcr: Also, if this is not likely that we add more things to `Request`, I would find it better to try keeping this interface simple and not use a new struct: `Process(req Request, addr Address) (resp proto.Message, err error)`, as you are already doing with Receiver.Recv

Gilthoniel: As discussed, I think it will be useful later.

blockchain/mod.go
```diff
@@ -29,10 +31,14 @@ type BlockFactory interface {
 	FromVerifiable(src proto.Message) (Block, error)
 }
 
-// Validator is the interface to implement to validate the generic payload
-// stored in the block.
-type Validator interface {
+// PayloadProcessor is the interface to implement to validate the generic
+// payload stored in the block.
+type PayloadProcessor interface {
```
nkcr: makes me think: we should also add the definition of "payload" to our terminologies.

blockchain/viewchange/mod.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"go.dedis.ch/fabric/blockchain"
+	"go.dedis.ch/fabric/mino"
+)
+
+// ViewChange provides primitives to verify if a participant is allowed to
+// propose a block as the leader. It is also responsible for verifying the
+// integrity of the players of the chain.
+type ViewChange interface {
+	// Wait returns a non-nil error when the node is allowed to make the
```
nkcr: - You return an error when you are allowed to make a proposal?
- I would use the same word in the comment and the code: in the comment you speak about "proposal" and in the code its "block". 
- It feels strange to suddenly speak about a "node". I would be a bit more precise by saying maybe "...returns a non-nil error when the node *running the code* is..."

mino/option.go
```diff
@@ -30,6 +30,25 @@ func ApplyFilters(filters []FilterUpdater) *Filter {
 // FilterUpdater is a function to update the filters.
 type FilterUpdater func(*Filter)
 
+// RotateFilter is a filter to rotate the indices. When n is above zero, it will
```
nkcr: so we should update the comment at line 12: "This list is always sorted."

blockchain/viewchange/mod.go
Gilthoniel: Oops, no it's a nil error when allowed. I changed node by player to be more consistent.

mino/option.go
Gilthoniel: There's already the comment ;)

PR #26, opened 2020-03-25
Skipchain: multiple inits create only one genesis block

Gilthoniel:


blockchain/skipchain/mod.go
```diff
@@ -172,8 +172,34 @@ func (a skipchainActor) InitChain(data proto.Message, players mino.Players) erro
...
+	}
+
+	if !xerrors.Is(err, NewNoBlockError(0)) {
+		return xerrors.Errorf("couldn't read the genesis block: %v", err)
+	}
+
 	conodes := newConodes(ca)
+	iter := conodes.AddressIterator()
+
+	if conodes.Len() > 0 && iter.GetNext().Equal(a.mino.GetAddress()) {
```
nkcr: what if `GetNext()` returns nil?

Gilthoniel: That's I check the length before

nkcr: Oh, right, I didn't pay attention. 
In that case that would make more sense using `iter.HasNext()` instead of ` conodes.Len() > 0`

Gilthoniel: Oops, yes. That's better.

PR #27, opened 2020-03-25
Initial ledger implementation

Gilthoniel:
This adds a initial skeleton for the Byzcoin ledger implementation.


ledger/byzcoin/mod.go
```diff
@@ -0,0 +1,249 @@
...
+		// TODO: closing
+		case rumor := <-ldgr.gossiper.Rumors():
+			tx, ok := rumor.(Transaction)
+			if ok {
+				ldgr.queue.Add(tx)
+			}
+		case <-roundTimeout:
+			// This timeout has two purposes. The very first use will determine
+			// the round time before the first block is proposed after a boot.
+			// Then it will be used to insure that blocks are still proposed in
```
nkcr: "... in *case* of ..."

ledger/byzcoin/mod.go
```diff
@@ -0,0 +1,249 @@
...
+			// of catastrophic failure in the consensus layer (i.e. too many
+			// players offline for a while).
+			go ldgr.proposeBlock(actor, players)
+
+			roundTimeout = time.After(timeoutRoundTime)
+		case block := <-blocks:
+			payload := block.GetPayload().(*BlockPayload)
```
nkcr: no check? (same in L.188)

ledger/byzcoin/mod.go
```diff
@@ -0,0 +1,249 @@
...
+
+func (ldgr *Ledger) makePayload(txs []Transaction) (*BlockPayload, error) {
+	payload := &BlockPayload{
+		Txs: make([]*TransactionProto, 0, len(txs)),
+	}
+
+	for _, tx := range txs {
+		packed, err := tx.Pack()
+		if err != nil {
+			return nil, err
```
nkcr:
```suggestion
			return nil, xerrors.Errorf("failed to pack tx: %v", err)
```

ledger/byzcoin/mod.go
```diff
@@ -0,0 +1,249 @@
...
+	// if it will be accepted.
+	err = actor.Store(payload, players)
+	if err != nil {
+		fabric.Logger.Err(err).Msg("couldn't send the payload")
+	}
+}
+
+func (ldgr *Ledger) makePayload(txs []Transaction) (*BlockPayload, error) {
+	payload := &BlockPayload{
+		Txs: make([]*TransactionProto, 0, len(txs)),
```
nkcr: what is the advantage over using `make([]*TransactionProto, len(txs))` with
```
for i, tx := range txs {
  payload.Txs[I] = ...
}
``` 
?

ledger/byzcoin/mod.go
```diff
@@ -0,0 +1,249 @@
...
+		for {
+			block, ok := <-blocks
+			if !ok {
+				fabric.Logger.Debug().Msg("watcher is closing")
+				return
+			}
+
+			payload := block.GetPayload().(*BlockPayload)
+
+			for _, pb := range payload.GetTxs() {
```
nkcr: if my understanding is correct, I would find it easier to follow if `pb` were renamed to `tx` and `tx` to `txProto`

ledger/byzcoin/mod.go
```diff
@@ -0,0 +1,249 @@
...
+		gossiper: g,
+	}
+}
+
+// AddTransaction implements ledger.Actor. It sends the transaction towards the
+// consensus layer.
+func (a actor) AddTransaction(in ledger.Transaction) error {
+	tx, ok := in.(Transaction)
+	if !ok {
+		return xerrors.Errorf("invalid message type '%T'", in)
```
nkcr:
```suggestion
		return xerrors.Errorf("invalid message type (expected "+
			"'ledger.Transaction', got '%T')", in)
```

mino/gossip/flat.go
```diff
@@ -0,0 +1,116 @@
...
+
+	"github.com/golang/protobuf/proto"
+	"github.com/golang/protobuf/ptypes"
+	"go.dedis.ch/fabric/encoding"
+	"go.dedis.ch/fabric/mino"
+	"golang.org/x/xerrors"
+)
+
+// Flat is an implementation of a message passing protocol that is using a flat
+// communication approach.
```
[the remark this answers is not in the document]
Gilthoniel: Oops, I forgot to tell you that it's a very initial implementation. The gossiper has been done quickly and is not yet polished. I'll update the comment anyway !

ledger/mod.go
```diff
@@ -0,0 +1,44 @@
...
+type Actor interface {
+	AddTransaction(tx Transaction) error
+}
+
+// Ledger provides the primitives to update a distributed public ledger through
+// transactions.
+type Ledger interface {
+	GetTransactionFactory() TransactionFactory
+
+	Listen(mino.Players) (Actor, error)
```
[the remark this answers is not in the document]
Gilthoniel: We should stay consistent with the previous modules that have the Listen already. My idea there was to the usual http server Listen as the function should start the different network capabilities of the module.

ledger/byzcoin/transaction.go
```diff
@@ -0,0 +1,146 @@
...
+	fmt "fmt"
+	"hash"
+
+	proto "github.com/golang/protobuf/proto"
+	"go.dedis.ch/fabric/crypto"
+	"go.dedis.ch/fabric/ledger"
+	"golang.org/x/xerrors"
+)
+
+// Digest is the type of a transaction ID.
```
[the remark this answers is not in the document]
Gilthoniel: Yes, I'm ending up using it in every module but I'm not sure it's a good idea to define it again and again. I guess the question how much we want to define an ID. We can define the length and let the hash algorithm as a choice or keep using a slice of bytes. It would be nice to have a chat about that at some point. For the moment I will keep it that way.

ledger/byzcoin/transaction.go
```diff
@@ -0,0 +1,146 @@
...
+func newTransactionFactory() transactionFactory {
+	return transactionFactory{
+		hashFactory: sha256Factory{},
+	}
+}
+
+// Create implements ledger.TransactionFactory. It returns a new transaction
+// built from the arguments.
+func (f transactionFactory) Create(args ...interface{}) (ledger.Transaction, error) {
+	tx, err := newTransaction(f.hashFactory, args[0].(string))
```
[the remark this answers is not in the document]
Gilthoniel: This will change later on, that was just to have a minimalistic implementation.

ledger/byzcoin/transaction.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	return tx, nil
+}
+
+// FromProto implements ledger.TransactionFactory. It returns a new transaction
+// built from the protobuf message.
+func (f transactionFactory) FromProto(pb proto.Message) (ledger.Transaction, error) {
+	var msg *TransactionProto
+	switch in := pb.(type) {
```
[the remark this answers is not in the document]
Gilthoniel: Your guess is right ;)

ledger/byzcoin/mod.go
Gilthoniel: None in that case, the parameter was previously a map and I forgot to update that part ;)

ledger/byzcoin/mod.go
Gilthoniel: Only `pb` to `txProto` but you're right.

Gilthoniel:
@nkcr Thanks for the comments! I'm sorry I forgot to precise it is an initial thus not everything polished.

PR #28, opened 2020-03-26
Creates new stream connection when an address is in the "neighbours" list

nkcr:
Now instead of always asking the orchestrator to relay messages, each server has a routing table and a "mesh" list, which is a list of server it must create a connection to. The routing table tells a server that doesn't have a direct connection to a peer which peer it can contact to relay that message.

I also implemented an optional way of saving all the trafic in a server. This is convenient in the tests to check that everything went well. It also provides interesting insights on how things work.


go.mod
```diff
@@ -8,7 +8,9 @@ require (
 	github.com/rs/zerolog v1.18.0
 	github.com/stretchr/testify v1.5.1
 	go.dedis.ch/kyber/v3 v3.0.12
+	golang.org/x/lint v0.0.0-20190313153728-d0100b6bd8b3 // indirect
```
Gilthoniel: Do those lines disappear with a `go mod tidy` ? My Go tools remove them all the time but I'm using 1.14 so they might have change something from 1.13

mino/minogrpc/mod.go
```diff
@@ -129,11 +129,11 @@ func (m Minogrpc) MakeRPC(name string, h mino.Handler) (mino.RPC, error) {
 	URI := fmt.Sprintf("%s/%s", m.namespace, name)
 	rpc := RPC{
 		handler: h,
-		srv:     m.server,
+		srv:     &m.server,
```
Gilthoniel: Should the server also be a pointer in the struct ?

mino/minogrpc/overlay.go
```diff
@@ -134,12 +171,12 @@ func (o overlayService) Stream(stream Overlay_StreamServer) error {
 			receiver.in <- msg
 		}
 	}()
-
+	o.handlerRcvr[apiURI[0]] = receiver
```
Gilthoniel: I don't think this should exist. Everything should happen in the handler.Stream function.

mino/minogrpc/server.go
```diff
@@ -149,25 +159,37 @@ func (rpc RPC) Stream(ctx context.Context,
...
 	}
 
 	// Creating a stream for each provided addr
 	for i := 0; players.AddressIterator().HasNext(); i++ {
 		addr := players.AddressIterator().GetNext()
 
-		clientConn, err := rpc.srv.getConnection(addr.String())
+		peer, ok := rpc.srv.neighbours[addr.String()]
+		if !ok {
+			err := xerrors.Errorf("addr '%s' not is our list of neighbours", addr)
```
Gilthoniel: You probably mean something else.

mino/minogrpc/server.go
```diff
@@ -441,50 +463,59 @@ func (s sender) Send(msg proto.Message, addrs ...mino.Address) <-chan error {
...
+		if neighbourFound {
+			clientConn, err := getConnection(addr.String(), peer, *s.srvCert)
+			if err != nil {
+				return xerrors.Errorf("failed to create new client in send: %v", err)
+			}
+			cl := NewOverlayClient(clientConn)
 
-		fabric.Logger.Trace().Msgf("sending back a message that must be "+
-			"relayed. From '%s', To '%s'", s.address.String(), addr.String())
+			header := metadata.New(map[string]string{headerURIKey: "blabla"})
```
Gilthoniel: What is this _blabla_ ?

mino/minogrpc/overlay.go
Gilthoniel: After reading, this should definitely be removed because a _Receiver_ is only an instance provided to the Stream handler, but it *must not* be transmitted. It has a given context and should close when this context is over. The same api URI must be able to manage several stream requests.

mino/minogrpc/overlay.go
```diff
@@ -17,7 +18,19 @@ import (
 // Server.Handlers, which is updated each time the makeRPC function is called.
 type overlayService struct {
 	handlers map[string]mino.Handler
-	addr     address
+	// this one is used when we open a new connexion on an existing RPC stream.
+	// In that case we must use the already existing receiver for this RPC.
+	handlerRcvr map[string]receiver
```
Gilthoniel: As mentionned below, this field should not exist.

mino/minogrpc/overlay.go
```diff
@@ -89,35 +102,59 @@ func (o overlayService) Stream(stream Overlay_StreamServer) error {
 			"header. Expected 1, found %d", len(apiURI))
 	}
 
+	rcvr, ok := o.handlerRcvr[apiURI[0]]
+	if ok {
+		go func() {
```
Gilthoniel: This logic should be separated from the receiver. A receiver transmits the message addressed to the node, it announces the end of the stream, etc .. Everything that a client should know.

This logic is to relay a message to a different node which is a totally different meaning.

mino/minogrpc/server.go
```diff
@@ -513,9 +545,11 @@ func sendSingle(s sender, msg proto.Message, addr mino.Address) error {
 
 // receiver implements mino.receiver
 type receiver struct {
-	errs chan error
-	in   chan *OverlayMsg
-	name string
+	errs    chan error
+	in      chan *OverlayMsg
+	name    string
+	stop    chan interface{}
```
Gilthoniel: The general advice is to use `chan struct{}` when you don't care about the type.

go.mod
nkcr: Updating from 1.12 to 1.14 solved the problem 👍

nkcr:
@Gilthoniel made the changes as discussed. You may want to first have a look at this test in order to check the logic:
https://github.com/dedis/fabric/blob/dfc4afe8f03e7d1721fb6e384dbe8bb107bbb7b6/mino/minogrpc/server_test.go#L962

mino/minogrpc/server_test.go
```diff
@@ -785,6 +790,327 @@ func TestRPC_MultipleRing_Stream(t *testing.T) {
...
+	require.Equal(t, "dummy_value_"+identifier2, msgs[1])
+	require.Equal(t, "dummy_value_"+identifier3, msgs[2])
+	require.Equal(t, "dummy_value_"+identifier4, msgs[3])
+
+	// fmt.Println("server1: " + server1.traffic.String())
+	// fmt.Println("server2: " + server2.traffic.String())
+	// fmt.Println("server3: " + server3.traffic.String())
+	// fmt.Println("server4: " + server4.traffic.String())
+
+	// // traffic of server 1
```
Gilthoniel: You forgot to uncomment ?

mino/minogrpc/traffic.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+type item struct {
+	typeStr string
+	addr    mino.Address
+	msg     proto.Message
+	context string
+}
+
+func (p item) String() string {
```
Gilthoniel: Could we define the `String` function to be a debug/log function that is minimalistic and one-liner (because `%v` is using it) and define a `Display(io.Writer)` function (in encoding for instance) that displays all the details like below ? So you can simply do `item.Display(os.Stdout)` ?

mino/minogrpc/server_test.go
```diff
@@ -968,6 +1293,36 @@ func (m fakeMembership) Len() int {
...
 }
 
+// Handler:
+// implements a handler where the stream sends back the message with its id
+type testMeshHandler struct {
+	mino.UnsupportedHandler
+	addrID string
+}
+
+func (t testMeshHandler) Stream(out mino.Sender, in mino.Receiver) error {
```
Gilthoniel: To make your test easier, you could have a wait group that is done when receiving a message so you don't have to manage the response in the test as you only want to verify that everyone is receiving the message.

Gilthoniel:
@nkcr I don't if the mesh/routing tables are temporary and you planned to improve this later so if that's the case, you can ignore the comments related to it.

nkcr:
Yes the routing/mesh structs are only temporary solutions. I agree this should be done in a *per request* fashion.

mino/minogrpc/overlay.go
```diff
@@ -95,43 +109,106 @@ func (o overlayService) Stream(stream Overlay_StreamServer) error {
 			"of handlers, did you register it?", apiURI[0])
 	}
 
+	rpcID := "server_" + o.addr.String()
```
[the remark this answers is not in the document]
nkcr: thing is the client and its corresponding RPC have the same address. So in this case the client will have "127.0.0.1:2001" and the RPC "server_127.0.0.1:2001".

mino/minogrpc/server_test.go
nkcr: got some trouble with having deterministic behaviour, making checking the logs impossible.
I'm wondering if we should have a flag that disable asynchronous calls: That would allow better testing and debugging.

mino/minogrpc/traffic.go
nkcr: yes, good idea

mino/minogrpc/server_test.go
nkcr: that's a solution. In this case my approach was that checking for a return is harder and tests more logic than only checking clients received the message.

PR #38, opened 2020-04-22
Implements a simple tree networking topology

nkcr:
As discussed, each node gets the list of addresses and computes in a deterministic manner the tree topology.

Tests will follow.


mino/minogrpc/overlay.go
```diff
@@ -110,6 +108,27 @@ func (o overlayService) Stream(stream Overlay_StreamServer) error {
...
+	err = o.encoder.UnmarshalAny(overlayMsg.Message, routingMsg)
+	if err != nil {
+		return xerrors.Errorf("failed to decode first routing message: %v", err)
+	}
+
+	addrs := make([]mino.Address, len(routingMsg.Addrs))
+	for i, addrStr := range routingMsg.Addrs {
+		addrs[i] = address{addrStr}
+	}
+	routing, err := NewTreeRouting(addrs, o.addr, treeeHeight)
```
Gilthoniel, line 127: `treeHeight`, also: can you add a TODO to allow a different routing algorithm ?

mino/minogrpc/overlay_test.go
```diff
@@ -191,3 +197,15 @@ func (t testFailHandler2) Stream(out mino.Sender, in mino.Receiver) error {
 	require.EqualError(t.t, err, "")
 	return nil
 }
+
+type goodEncoder struct {
```
Gilthoniel, line 201: you can use `encoding.ProtoEncoder` that implements encoding.ProtoMarshaler instead of the fake ones. Then keep your function implementation to ignore.

mino/minogrpc/routing.go
```diff
@@ -0,0 +1,201 @@
...
+// NewTreeRouting creates the network tree in a deterministic manner based on
+// the list of addresses. Please be careful that the provided list of addresses
+// is shuffled, which could affect subsequent use of this list, especially
+// outside the function's scope.
+func NewTreeRouting(addrs []mino.Address, addr mino.Address,
+	treeHeight int) (*TreeRouting, error) {
+
+	addrsStr := make([]string, len(addrs))
+	for i, addr := range addrs {
+		addrsStr[i] = addr.String()
```
Gilthoniel, line 36: Use `addr.MarshalText()` instead as you don't know if the string representation is biased.

Gilthoniel: To sort, you can define a slice type then use `bytes.Compare`.

mino/minogrpc/routing.go
```diff
@@ -0,0 +1,201 @@
...
+	hash := sha256.New()
+	for _, addr := range addrsStr {
+		_, err := hash.Write([]byte(addr))
+		if err != nil {
+			fabric.Logger.Fatal().Msgf("failed to write hash: %v", err)
+		}
+	}
+
+	hashBuf := bytes.NewBuffer(hash.Sum(nil))
+	hashInt, err := binary.ReadVarint(hashBuf)
```
Gilthoniel, line 50: It's easier to implement this in a different language if it is defined properly like `binary.LittleEndian.ReadUInt64`. Because you have to look for the Varint way of reading.

Gilthoniel: As you're using sha256, you know the size.

Gilthoniel:
We can have a student project next semester to implement a more resilient tree.

mino/minogrpc/routing.go
nkcr: something like `binary.LittleEndian.Uint64(hash.Sum(nil))` ?

Gilthoniel: yep that one

nkcr:
>We can have a student project next semester to implement a more resilient tree.

Yes, that would be a well-defined and interesting project. In that sense I replaced the treeRouting by an interface. 

I also updated the test to check if the mesh is fully connected. For the record, here is the complete ordered traffic when 9 nodes relay a packet in a ring fashion with the following sequence `client -> N1 -> N2 -> N3 ... N9 -> client`

graphviz

nkcr:
@Gilthoniel ready for a second review. I still have some deadlocks when I play with >15 nodes with DKG, but I'm waiting to merge this PR and will continue investigating in #41 . But maybe you'll already discover something suspicious there. Another thing: can you have a particular look at how contexts are used? I realised I was misusing them but I don't know if its 100% ok now.

mino/minogrpc/server_test.go
```diff
@@ -1299,46 +1235,123 @@ func (t testMeshHandler) Stream(out mino.Sender, in mino.Receiver) error {
...
+		Address:     server.listener.Addr().String(),
+		Certificate: server.cert.Leaf,
+	}
+
+	return identifier, addr, server, peer
+}
+
+// Sleep random
+
+func rsleep() {
```
Gilthoniel, line 1354: I'm not a big fan of the random sleep. You expect from a unit test to make sure a particular scenario will end as expected. You manually send messages in different order to test that they will be handled as expected. Random sleep only provokes random errors on Travis.

mino/minogrpc/traffic.go
```diff
@@ -6,32 +6,44 @@ import (
...
+	"go.dedis.ch/fabric/encoding"
 	"go.dedis.ch/fabric/mino"
 )
 
 var eachLine = regexp.MustCompile(`(?m)^(.+)$`)
 
+var counter = atomicCounter{}
+
 // traffic is used to keep track of packets traffic in a server
 type traffic struct {
```
Gilthoniel, line 21: Can you either test this file, or move it to an internal/testing package ?

mino/minogrpc/mod.go
```diff
@@ -129,13 +128,21 @@ func (m Minogrpc) MakeNamespace(namespace string) (mino.Mino, error) {
...
+		routingFactory: m.server.routingFactory,
 	}
 
 	m.server.handlers[URI] = h
 
 	return rpc, nil
 }
+
+// AddNeighbours fills the neighbours map of the server
+func (m Minogrpc) AddNeighbours(minoGrpcs ...*Minogrpc) {
```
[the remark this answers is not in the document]
nkcr: yes you're right, it belongs to a utility function in the test,

mino/minogrpc/mod_test.go
```diff
@@ -16,6 +16,7 @@ func TestMessages(t *testing.T) {
 	messages := []proto.Message{
 		&OverlayMsg{},
 		&Envelope{},
```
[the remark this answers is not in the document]
nkcr: This is something I should merge, but right now it would just add a new layer of uncertainty. I'm adding it to my dodo list

mino/minogrpc/overlay.go
```diff
@@ -122,24 +144,29 @@ func (o overlayService) Stream(stream Overlay_StreamServer) error {
 		// doesn't relay but sends directly.
 		address:      address{rpcID},
```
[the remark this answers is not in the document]
nkcr: mmmm it is indeed the address of the sender, which there acts as the orchestrator for its children. But from the point of view of the children yes, this address acts as a gateway.

mino/minogrpc/overlay.go
```diff
@@ -177,49 +206,45 @@ func (o overlayService) Stream(stream Overlay_StreamServer) error {
...
+		}(addr)
 	}
 
 	// listen on my own stream
 	go func() {
+
 		for {
 			err := listenStream(stream, &receiver, sender, o.addr)
 			if err == io.EOF {
+				<-ctx.Done()
```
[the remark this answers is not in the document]
nkcr: I'm not sure either, I added it during my deadlock chase.  If that's ok like that I'll remove that when I will be sure it still works against the DKG test in the other PR. Added in the TODO.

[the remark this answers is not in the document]
nkcr: Thing is I can't exist once the handler is done, because a node has two responsibilities: running the handler and relaying messages. So if I exist once the handler is done, then the node won't relay messages anymore, which can be problematic.

mino/minogrpc/server_test.go
nkcr: Yes I agree. I would however find interesting at some point to have some sort of fuzzing tests with different timing to stress the system. But of course not in this context.

mino/minogrpc/traffic.go
nkcr: Yes testing is planed. Added in the TODO because for now this is still very prone to change.

mino/minogrpc/overlay.proto
```diff
@@ -18,4 +18,9 @@ service Overlay {
 
 message OverlayMsg {
     google.protobuf.Any message = 1;
+}
+
+message RoutingMsg {
```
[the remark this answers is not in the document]
nkcr: Based on this comment I refactored the routing interface in its own package. The Routing now implements `encoding.Packable` and the factory has a `FromAny(*any.Any)`.

nkcr:
Thanks @Gilthoniel for the review. I addressed your comments. The only thing that would need a second pass would be the refactor of the routing interface, which is now in `minogrpc/routing`. (Tests for this package will be added later.)

mino/minogrpc/routing/routing.go
```diff
@@ -0,0 +1,428 @@
...
+
+	if len(addrsBuf) == 0 {
+		return nil, xerrors.Errorf("there should be at least one address in " +
+			"the iterator")
+	}
+
+	sort.Stable(&addrsBuf)
+
+	// We will use the hash of the addresses to set the random seed.
+	hash := sha256.New()
```
Gilthoniel, line 117: the two functions have a lot of code repetition that could into a private function.

nkcr: done 👍

PR #41, opened 2020-04-27
Adds a first simple DKG implementations

nkcr:
Adds a first DKG/Pedersen implementation and improves minogrpc.

- Defines a DKG interface in `dkg/`
- Implements Pedersen in `dkg/pedersen`
- Updates minogrpc to only use Enveloppe
- Updates and fixes the tree routing protocol
- Fixes some race conditions with minogrpc
- Updates minogrpc/traffic to display a more accurate representation

More extensive testing will follow.


nkcr:
For the record, here is an update of the networking flow diagram for the `TestRPC_MultipleRingMesh_Stream` test where messages flow as: 

```
CLIENT -> NODE1 -> NODE2 -> ... NODE9 -> CLIENT
```

tree height is set to 4.

graph

dkg/pedersen/handler.go
```diff
@@ -0,0 +1,386 @@
...
+		h.RLock()
+		decryptReply := &DecryptReply{
+			V: VBuf,
+			// TODO: check if using the private index is the same as the public
+			// index.
+			I: int64(h.privShare.I),
+		}
+		h.RUnlock()
+
+		errs := out.Send(decryptReply, from)
```
Gilthoniel, line 107: `err = <-out.Send(...)` as you're listening to one error anyway, so either it's nil or something went wrong.

dkg/pedersen/handler.go
```diff
@@ -0,0 +1,386 @@
...
+				"reply: %v", err)
+		}
+
+	case *Deal:
+		// This is a special case where a DKG started, some nodes received the
+		// start signal and started sending their deals but we have not yet
+		// received our start signal. In this case we will collect the Deals
+		// while waiting for the start signal.
+		deals := []*Deal{msg}
+		for {
```
Gilthoniel, line 120: You're essentially doing the same thing so you can probably for loop the whole select and remove this one.

Gilthoniel: You just need to handle the _Decrypt_ if it hasn't started but you have to do it anyway.

dkg/pedersen/handler.go
```diff
@@ -0,0 +1,386 @@
...
+			fabric.Logger.Warn().Msgf("%s failed to process response "+
+				"from '%s': %v", h.me, from, err)
+		}
+	}
+
+	for !h.dkg.Certified() {
+
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+		from, msg, err := in.Recv(ctx)
```
Gilthoniel, line 272: In that case, you expect the context to have some timeout if one of the N-1 node cannot answer for instance. Also the cancel is only triggered at the end of the function so it's not really useful here. The context is there to stop listening sooner if something happens. Here it seems to me a timeout is the only event that could happen.

dkg/pedersen/handler.go
```diff
@@ -0,0 +1,386 @@
...
+				Signature: deal.Deal.Signature,
+				Nonce:     deal.Deal.Nonce,
+				Cipher:    deal.Deal.Cipher,
+			},
+			Signature: deal.Signature,
+		}
+
+		errs := out.Send(dealMsg, addrs[i])
+		go func(errs <-chan error) {
+			err, more := <-errs
```
Gilthoniel, line 195: `err = <-errs` if only is expected

dkg/pedersen/handler.go
```diff
@@ -0,0 +1,386 @@
...
+
+	for _, addr := range addrs {
+		if addr.Equal(h.me) {
+			wg.Done()
+			continue
+		}
+
+		errs := out.Send(respProto, addr)
+		go func(errs <-chan error) {
+			err, more := <-errs
```
Gilthoniel, line 375: `err = <-errs` if only one is expected

dkg/pedersen/mod.go
```diff
@@ -0,0 +1,258 @@
...
+type starter struct {
+	pubKeys []kyber.Point
+	privKey kyber.Scalar
+	m       mino.Mino
+	suite   suites.Suite
+	rpc     mino.RPC
+}
+
+// Start implements dkg.Starter. It allows one to initialize a new DKG protocol.
+func (s *starter) Start(players mino.Players, t uint32) (dkg.DKG, error) {
```
Gilthoniel, line 77: The public keys should probably be provided at that point. I guess you need to see with Calypso how you're going to manage the keys but they should basically be stored on the blockchain when integrated to Byzcoin.

dkg/pedersen/mod_test.go
```diff
@@ -0,0 +1,128 @@
...
+	"go.dedis.ch/fabric/mino/minoch"
+	"go.dedis.ch/fabric/mino/minogrpc"
+	"go.dedis.ch/fabric/mino/minogrpc/routing"
+	"go.dedis.ch/kyber/v3"
+)
+
+func TestPedersen_Scenario(t *testing.T) {
+	n := 10
+
+	addrFactory := minoch.AddressFactory{}
```
Gilthoniel, line 19: Why are you using this ? It should work with the `fake.Address` like skipchain and byzcoin.

mino/minogrpc/mod.go
```diff
@@ -143,3 +147,10 @@ func (m Minogrpc) MakeRPC(name string, h mino.Handler) (mino.RPC, error) {
 
 	return rpc, nil
 }
+
+// AddNeighbours fills the neighbours map of the server
+func (m Minogrpc) AddNeighbours(minoGrpcs ...*Minogrpc) {
```
Gilthoniel, line 152: You'll have to change this anyway when going to real world, so maybe it's the right time to make an API to pass the certificate?

mino/minogrpc/overlay.go
```diff
@@ -177,18 +184,30 @@ func (o overlayService) Stream(stream Overlay_StreamServer) error {
...
-		sender.participants[addr.String()] = clientStream
+		clientStream := newSafeOverlayStream(cs)
+
+		sender.participants.Store(addr.String(), clientStream)
 
 		// Sending the routing info as first messages to our children
-		clientStream.Send(&OverlayMsg{Message: overlayMsg.Message})
+		clientStream.Send(&Envelope{
+			From:         o.addr.String(),
+			PhysicalFrom: o.addr.String(),
```
Gilthoniel, line 202: I would use something more meaningful like `LastHop` or `Relay`. Now, is it really necessary ? You're supposed to know the mesh.

mino/minogrpc/overlay.go
```diff
@@ -177,18 +184,30 @@ func (o overlayService) Stream(stream Overlay_StreamServer) error {
...
+			To:           []string{addr.String()},
+			Message:      envelope.Message,
+		})
+
+		// TODO: think how we can make all the node set the routing then start
```
Gilthoniel, line 207: It should be impossible to receive a message before the routing message, because the relay will open the connection and then directly send the routing message before relaying anything.

mino/minogrpc/server.go
```diff
@@ -400,6 +401,15 @@ func (srv *Server) Serve() error {
 	return nil
 }
 
+func (srv *Server) addNeighbour(servers ...*Server) {
+	for _, server := range servers {
+		srv.neighbours[server.addr.String()] = Peer{
```
Gilthoniel, line 406: As mentioned earlier, at some point you will need to be able to extract the server certificate (to a file for instance), transmit it to a peer and then populate the peer server with it. From what I see, it shouldn't hard to adapt but as soon as it's done, you won't get more surprises.

mino/minogrpc/server.go
```diff
@@ -530,28 +544,33 @@ func (s *sender) sendSingle(msg proto.Message, from, to mino.Address) error {
...
-		From:    from.String(),
-		To:      []string{to.String()},
-		Message: msgAny,
-	}
-	envelopeAny, err := s.encoder.MarshalAny(envelope)
-	if err != nil {
-		return xerrors.Errorf("couldn't marshal envelope: %v", err)
+	player, ok := playerItf.(overlayStream)
+	if !ok {
+		fabric.Logger.Fatal().Msg("not ok")
```
Gilthoniel, line 562: I can't believe I'm the one saying an error message is too short ! But it should slightly more detailed. Why would that happen ?

Gilthoniel:
When you implement the tests, if you're using minogrpc, it would be nice to make it real in the sense that you don't have access to other servers. I do that with minoch because that's the purpose but a few integration tests with minogrpc are definitely nice.

dkg/pedersen/mod_test.go
Gilthoniel: Actually, the address should come from the instance of Mino. When you instantiate an implementation of Mino, it creates its own address and you should get it from there.

Gilthoniel:
```
m := minogrpc.NewMinogrpc("0.0.0.0", 2000, treeFactory) // it should be done like that
addrs[i] = m.GetAddress()
```

dkg/mod.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+// Factory defines the DKG Starter's factory
+type Factory interface {
+	New(pubKeys []kyber.Point, privKey kyber.Scalar, m mino.Mino,
+		suite suites.Suite) (Starter, error)
+}
+
+// Starter defines the primitive to start a DKG protocol
+type Starter interface {
```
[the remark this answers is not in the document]
nkcr: I'm always for keeping consistency. And yes, it makes sense!
Like so?

```go
// DKG defines the primitive to start a DKG protocol
type DKG interface {
	Listen(players mino.Players, threshold uint32) (DKG, error)
}

// Actor defines the primitives to use a DKG protocol
type Actor interface {
	Decrypt(K, C kyber.Point) ([]byte, error)
	Reshare() error
}
```

mino/minogrpc/server.go
nkcr: Ugh, you're so right! I forgot to update this one.... : /

mino/minogrpc/overlay.go
nkcr: IMO it can't be more explicit than that. `LastHop` is already speaking at a logical level, where here I want to know who sent me this message, regardless its content.
It is hard to know for a node who sent a message because the listen function receive messages from the orchestrator and the children of a node.

nkcr:
>  it would be nice to make it real in the sense that you don't have access to other servers

Yes that's is already the case in the `server_test.go` except when I can specify a dynamic number of servers (which happens once).

PR #42, opened 2020-05-01
Byzcoin roster changes

Gilthoniel:
This PR refactors the transactions in Byzcoin to allow multiple kinds to take advantage of the polymorphism. It then partially implements the roster transactions.

Coming: Byzcoin unit tests and full roster change support.


Gilthoniel:
@nkcr The PR seems huge because I split the transaction abstraction and the task abstraction (equivalent to an instruction in cothority). Most of the new code is the task implementations for DARC, contracts and the roster.

ledger/byzcoin/mod.go
```diff
@@ -48,66 +46,58 @@ type Ledger struct {
...
 
-// GetInstance implements ledger.Ledger. It returns the instance with the given
-// key as of the latest block.
-func (ldgr *Ledger) GetInstance(key []byte) (consumer.Instance, error) {
+// GetValue implements ledger.Ledger.
+func (ldgr *Ledger) GetValue(key []byte) (proto.Message, error) {
 	latest, err := ldgr.bc.GetBlock()
 	if err != nil {
-		return nil, xerrors.Errorf("couldn't read latest block: %v", err)
+		return nil, err
```
nkcr, line 87: why removing those error wrappings?

ledger/byzcoin/task.go
```diff
@@ -0,0 +1,74 @@
...
+	"go.dedis.ch/fabric/encoding"
+	"go.dedis.ch/fabric/ledger/arc/darc"
+	"go.dedis.ch/fabric/ledger/byzcoin/roster"
+	"go.dedis.ch/fabric/ledger/inventory"
+	"go.dedis.ch/fabric/ledger/transactions/basic"
+	"go.dedis.ch/fabric/mino"
+	"golang.org/x/xerrors"
+)
+
+// taskFactory is an task factory that can process several types of tasks.
```
nkcr, line 19:
```suggestion
// taskFactory is a task factory that can process several types of tasks.
```

ledger/inventory/mod.go
```diff
@@ -12,7 +12,8 @@ type Page interface {
 	// the integrity.
 	GetFootprint() []byte
 
-	// Read returns the value stored at the given key.
+	// Read returns the value stored at the given key. If the key does not
+	// exist, it should a nil value without error.
```
nkcr, line 16:
```suggestion
	// exist, it should return a nil value without error.
```

encoding/mod.go
```diff
@@ -49,6 +49,12 @@ type ProtoMarshaler interface {
 	UnmarshalDynamicAny(any *any.Any) (proto.Message, error)
 }
 
+// Fingerprinter is an interface to perform fingerprinting on object.
+type Fingerprinter interface {
+	// Fingerprint writes itself to the writer in a deterministic way.
+	Fingerprint(io.Writer, ProtoMarshaler) error
```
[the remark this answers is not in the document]
Gilthoniel: Good point I forgot to change footprint to fingerprint which seems to be more used.

ledger/byzcoin/mod.go
Gilthoniel: Oh that's because I removed the GetInstance which does not really makes sense anymore and I usually write the errors alongside the tests but the tests will come in the PR.

PR #43, opened 2020-05-11
Finalization of the roster change

Gilthoniel:
- Implements a catch up process for new joining participant
- Implements roster change including new players
- Unit tests


ledger/byzcoin/roster/task.go
```diff
@@ -3,50 +3,81 @@ package roster
...
 func (t clientTask) GetChangeSet() viewchange.ChangeSet {
+	removals := append([]uint32{}, t.remove...)
+	// Sort by ascending order in O(n*log(n)).
+	sort.Slice(removals, func(i, j int) bool { return removals[i] > removals[j] })
+	// Remove duplicates in O(n).
+	for i := 0; i < len(removals)-1; {
+		if removals[i] == removals[i+1] {
+			removals = append(removals[:i], removals[i+1:]...)
+		} else {
+			// Only moves to the next when all occurances of the same index are
```
nkcr, line 69:
```suggestion
			// Only moves to the next when all occurences of the same index are
```

ledger/byzcoin/roster/task.go
```diff
@@ -247,12 +294,79 @@ func (f TaskManager) FromProto(in proto.Message) (basic.ServerTask, error) {
 		return nil, xerrors.Errorf("invalid message type '%T'", in)
 	}
 
+	player, err := f.unpackPlayer(pb.GetAddr(), pb.GetPublicKey())
+	if err != nil {
+		// The error is not wrap to avoid redundancy with the private function.
+		return nil, err
```
nkcr, line 300: I am not sure why this would be redundant.

For example without wrap you could have
```
couldn't decode public key
```

and if you add one:
```
failed to unpack player: couldn't decode public key
```

If I had to choose I would go for the second one. IMO no wrap should be an extreme exception where there is no other way around, because there is often a good reason to avoid a wrap but it is not in our interest when something bad happens. Having more and more exceptions is a slippery slope, (I feel this PR as an unusual number of no-wrap error).

Gilthoniel: I have the tendency to use private functions to reduce the code block sizes because I don't like big blocks but I assume it will be redundant when doing that. But it's true it's not always a benefit.

PR #47, opened 2020-05-19
Minogrpc refactoring

Gilthoniel:
This refactors minogrpc to reduce the amount of code. It also implements an infinite buffered receiver to prevent the relay from stalling. Finally, it adds unit tests.


mino/minogrpc/server.go
```diff
@@ -47,384 +38,289 @@ var (
...
-				continue
-			}
+	from := o.routingFactory.GetAddressFactory().FromText(msg.GetFrom())
 
-			resp, err := rpc.encoder.UnmarshalDynamicAny(callResp.Message)
-			if err != nil {
-				errs <- xerrors.Errorf("couldn't unmarshal message: %v", err)
-				continue
-			}
+	req := mino.Request{
```
nkcr, line 76: I am suddenly wondering: why don't we use the same type of return for streaming? Right now its 
```
Recv(context.Context) (Address, proto.Message, error)
```

but shouldn't it be 

```
Recv(context.Context) (Request, error)
```
 ?

mino/minogrpc/server.go
```diff
@@ -47,384 +38,289 @@ var (
...
+	// We fetch the uri that identifies the handler in the handlers map with the
+	// grpc metadata api. Using context.Value won't work.
+	headers, ok := metadata.FromIncomingContext(ctx)
+	if !ok {
+		return nil, xerrors.Errorf("header not found in provided context")
 	}
 
-	sendMsg := &Envelope{
-		Message: m,
+	apiURI := headers[headerURIKey]
```
nkcr, line 58:
```
uri := uriFromContext(ctx)
...
```

mino/minogrpc/server.go
```diff
@@ -47,384 +38,289 @@ var (
...
-// NewServer sets up a new server
-func NewServer(addr mino.Address, rf routing.Factory) (*Server, error) {
-	if addr.String() == "" {
-		return nil, xerrors.New("addr.String() should not give an empty string")
+// GetCertificate returns the certificate of the overlay.
+func (o overlay) GetCertificate() *tls.Certificate {
+	cert, found := o.certs.Load(o.me)
+	if !found {
+		// This should never and it will panic if it does as this will provoke
+		// several issues later on.
```
nkcr, line 190:
```suggestion
		// This should never happen and it will panic if it does as this will
		// provoke several issues later on.
```

mino/minogrpc/server.go
```diff
@@ -47,384 +38,289 @@ var (
...
+
 	return nil
 }
 
-// AddCertificate populates the list of public know certificates of the server
-func (srv *Server) addCertificate(key string, cert *x509.Certificate) {
-	srv.nodesCerts.Store(key, cert)
-}
+func (o overlay) setupStream(stream relayer, sender *sender, receiver *receiver, addr mino.Address) {
+	// Relay sender for that connection.
```
nkcr, line 269:
```suggestion
func (o overlay) setupStream(stream relayer, sender *sender, receiver *receiver,
	addr mino.Address) {

	// Relay sender for that connection.
```

mino/minogrpc/stream.go
```diff
@@ -0,0 +1,227 @@
...
+
+// Queue is an interface to queue messages.
+type Queue interface {
+	Channel() <-chan *Message
+	Push(*Message)
+}
+
+// NonBlockingQueue is an implementation of a queue that makes sure pushing a
+// message will never hang.
+type NonBlockingQueue struct {
```
nkcr, line 171: I would personally call it `InfiniteQueue`

mino/minogrpc/stream.go
```diff
@@ -0,0 +1,227 @@
...
+		if !q.running {
+			q.running = true
+			go q.pushAndWait()
+		}
+		q.Unlock()
+	}
+}
+
+func (q *NonBlockingQueue) pushAndWait() {
+	q.working.Add(1)
```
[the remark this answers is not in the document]
Gilthoniel: for now it's only used in the tests but the purpose is to make sure the go routine has stopped.

mino/minogrpc/mod.go
```diff
@@ -60,46 +97,129 @@ type AddressFactory struct{}
...
+	if len(text) == 0 {
+		return newRootAddress()
+	}
+
+	return address{host: string(text)}
 }
 
-// NewMinogrpc sets up the grpc and http servers. URL should
-func NewMinogrpc(serverURL *url.URL, rf routing.Factory) (*Minogrpc, error) {
+// Minogrpc represents a grpc service restricted to a namespace
```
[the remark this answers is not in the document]
Gilthoniel: Added to the struct definition.

mino/minogrpc/stream.go
Gilthoniel: The buffer should be limited at some point to limit the usage of memory so it actually shouldn't be infinite. I'll add a TODO for that.

mino/minogrpc/server.go
Gilthoniel: As discussed, will do after the PR.

PR #51, opened 2020-05-26
Adds a utility to generate the dependency graph of each package

nkcr:


internal/depgraph/mod.go
```diff
@@ -0,0 +1,324 @@
...
+		if !os.IsNotExist(err) && !c.Bool("force") {
+			return xerrors.Errorf("file '%s' already exist, use '-F' to "+
+				"overwrite", c.String("out"))
+		}
+
+		out, err = os.Create(c.String("out"))
+		if err != nil {
+			return xerrors.Errorf("failed to create output file: %v", err)
+		}
+	}
```
pierluca, line 165: The overall `run` function is extremely long.

I would extract one or more functions dedicated to parsing the configuration, hence improving readability.

internal/depgraph/mod.go
```diff
@@ -0,0 +1,324 @@
...
+		_, err := os.Stat(c.String("out"))
+		if !os.IsNotExist(err) && !c.Bool("force") {
+			return xerrors.Errorf("file '%s' already exist, use '-F' to "+
+				"overwrite", c.String("out"))
+		}
+
+		out, err = os.Create(c.String("out"))
+		if err != nil {
+			return xerrors.Errorf("failed to create output file: %v", err)
+		}
```
pierluca, line 164: this by itself could be a function returning `out, err`

internal/depgraph/mod.go
```diff
@@ -0,0 +1,324 @@
...
+	for _, it := range config.Interfaces {
+		interfaces[it] = struct{}{}
+	}
+
+	// links will contain, for every package, a bag of dependencies. The bag is
+	// done with a dummy map.
+	links := make(map[string]map[string]struct{})
+
+	// parseFile will be called recursively on each file and folder
+	parseFile := func(path string, f os.FileInfo, err error) error {
```
pierluca, line 178: same reasoning, to improve readability: 
- I would extract this as a separate function, taking config & links as an extra parameter
- call filepath.Walk with a simple function wrapper which inserts the parameters

internal/depgraph/mod.go
```diff
@@ -0,0 +1,324 @@
...
+
+		// we exclude the dir and non-go files
+		if f.IsDir() || !strings.HasSuffix(f.Name(), ".go") ||
+			strings.HasSuffix(f.Name(), "_test.go") {
+
+			return nil
+		}
+
+		// we exclude test files if not otherwise asked
+		if !c.Bool("withTest") && strings.HasSuffix(f.Name(), "_test.go") {
```
pierluca, line 191: I'd consider moving the "withTest" parameter to the overall configuration

pierluca:
Nice tool ! 

I don't know the context, so I may be missing something, but taking a step back, isn't there something out there we could have used (or adapted) to do the same job?

internal/depgraph/mod.go
nkcr: You mean the config file? Doing that means you must provide the config file to use it, where sometimes you don't want to do that. Since this is just a simple bool argument that shouldn't be used that often, I'd prefer keeping it as it is, like the "-F" argument.

pierluca: No, I meant the `type config struct`

nkcr:
Thanks for the review. Indeed I didn't find any tool that would suit our needs. There is the `go mod graph` that can somehow print a graph, but then we need to filter stuff and highlight our interfaces.
`depgraph` stays an internal tool for dela (internal/ is meant for that) that does a tailored job we need.

internal/depgraph/mod.go
```diff
@@ -0,0 +1,360 @@
...
+	"strings"
+	"time"
+
+	"github.com/urfave/cli/v2"
+	"golang.org/x/xerrors"
+	"gopkg.in/yaml.v2"
+)
+
+/*
+	This package provides a utility CLI to generate the dependency graph of each
```
Gilthoniel, line 23: Can you use the Go doc syntax ? At the beginning of the file with `Package main provides ...`

Gilthoniel: The first line is seen in the doc, and if you make a new line for the example, it's only when you go read the package doc.

internal/depgraph/mod.go
```diff
@@ -0,0 +1,360 @@
...
+	}
+
+	configPath := c.String("config")
+	if configPath != "" {
+		configBuf, err := ioutil.ReadFile(configPath)
+		if err != nil {
+			return config, xerrors.Errorf("failed to read config file: %v", err)
+		}
+
+		err = yaml.Unmarshal([]byte(configBuf), &config)
```
Gilthoniel, line 182: `err = yaml.Unmarshal(configBuf, &config)`

internal/depgraph/mod.go
```diff
@@ -0,0 +1,360 @@
...
+		return os.Stdout, nil
+	}
+
+	_, err := os.Stat(config.OutFile)
+	if !os.IsNotExist(err) && !config.OverWrite {
+		return nil, xerrors.Errorf("file '%s' already exist, use '-F' to "+
+			"overwrite", config.OutFile)
+	}
+
+	out, err := os.Create(config.OutFile)
```
Gilthoniel, line 207: If folders are given in the parameter and they don't exist, it will fail.

Gilthoniel (approved):

PR #54, opened 2020-05-27
Tentative of a cli template

Gilthoniel:
This PR implements a tentative of a CLI template for the modules to follow to implement their own controllers that can be chosen before compiling an ledger application.

Also: implement a controller to create a network of nodes with grpc


Gilthoniel:
example

With nodes started using: `memcoin --socket /tmp/node1.sock start --port 2000`

mino/minogrpc/controller/mod.go
```diff
@@ -0,0 +1,93 @@
...
+				Required: true,
+			},
+		).
+		Action(joinAction{})
+}
+
+// Run implements cmd.Controller. It starts the minogrpc instance and inject it
+// in the dependency resolver.
+func (m minimal) Run(ctx cmd.Context, inj cmd.Injector) error {
+	rf := routing.NewTreeRoutingFactory(3, minogrpc.AddressFactory{})
```
nkcr, line 76: we might want to extract `3` to a constant

mino/minogrpc/mod.go
```diff
@@ -97,16 +101,31 @@ type AddressFactory struct{}
...
+// Joinable is an extension of the mino.Mino interface to allow distant servers
+// to join a network of participants.
+type Joinable interface {
+	mino.Mino
+
+	GetCertificate() *tls.Certificate
+
+	GetCertificateStore() certs.Storage
+
+	Token(expiration time.Duration) string
```
nkcr, line 120: We use Get+element above, why not here? Or maybe `GenerateToken`.

mino/minogrpc/overlay.proto
```diff
@@ -4,6 +4,29 @@ package minogrpc;
 
 import "google/protobuf/any.proto";
 
+// Certificate is wrapper around a x509 raw certificate and its address.
```
nkcr, line 7:
```suggestion
// Certificate is a wrapper around a x509 raw certificate and its address.
```

mino/minogrpc/overlay.proto
```diff
@@ -4,6 +4,29 @@ package minogrpc;
...
+    bytes address = 1;
+    bytes value = 2;
+}
+
+message CertificateAck {}
+
+// JoinRequest sends a request to join a network to a distant node. It must
+// contain a valid token and its own certificate.
+message JoinRequest {
+    string token = 1;
```
nkcr, line 18: Is the token a hex encoded string or really any unicode char?

mino/minogrpc/overlay.proto
```diff
@@ -16,6 +39,14 @@ message Envelope {
 }
 
 service Overlay {
+    // Join handles join request from a unknown node. It accepts to share the
```
nkcr, line 42:
```suggestion
    // Join handles join request from an unknown node. It accepts to share the
```

mino/minogrpc/overlay.proto
```diff
@@ -16,6 +39,14 @@ message Envelope {
...
+    // certificates if the token is valid.
+    rpc Join(JoinRequest) returns (JoinResponse) {}
+
+    // Share handle a certificate share from another participant of the network.
```
nkcr, line 46:
```suggestion
    // Share handles a certificate share from another participant of the network.
```

mino/minogrpc/server.go
```diff
@@ -45,6 +47,83 @@ type overlayServer struct {
...
+
+	for to, cert := range list {
+		text, err := to.MarshalText()
+		if err != nil {
+			return nil, xerrors.Errorf("couldn't marshal address: %v", err)
+		}
+
+		msg := &Certificate{Address: text, Value: cert}
+
+		// Prepare the list of send back to the new node.
```
nkcr, line 74: it's too late for me to rephrase it, but it seems that a word should be changed.

mino/minogrpc/server.go
```diff
@@ -45,6 +47,83 @@ type overlayServer struct {
...
+		ack++
+	}
+
+	// 3. Return the set of known certificates.
+	return &JoinResponse{Peers: peers}, nil
+}
+
+func (o overlayServer) Share(ctx context.Context, msg *Certificate) (*CertificateAck, error) {
+	// TODO: verify the validity of the certificate by connecting to the distant
+	// node by that requires a protection against malicious share.
```
nkcr, line 113:
```suggestion
	// node but that requires a protection against malicious share.
```

mino/minogrpc/tokens/mod.go
```diff
@@ -0,0 +1,59 @@
+// Package tokens defines a token holder to generate and validate access tokens.
+// and provides an implementation in-memory.
```
nkcr, line 2:
```suggestion
// Package tokens defines a token holder to generate and validate access tokens
// and provides an in-memory implementation.
```

mino/minogrpc/tokens/mod.go
```diff
@@ -0,0 +1,59 @@
...
+
+	holder.Lock()
+	holder.tokens[str] = time.Now().Add(expiration)
+	holder.Unlock()
+
+	return str
+}
+
+// Verify implements tokens.Holder. It returns true if the token is valid.
+func (holder *InMemoryHolder) Verify(token string) bool {
```
nkcr, line 49: shouldn't the token also store the node address it has been issued to? Right now someone could potentially eave drop on the network, stole the token and use it. Verify would then be

```go
Verify(token string, origin mino.Address)
```

mino/minogrpc/server.go
```diff
@@ -319,6 +461,7 @@ func makeCertificate() (*tls.Certificate, error) {
 
 	tmpl := &x509.Certificate{
 		SerialNumber: big.NewInt(1),
+		DNSNames:     []string{},
```
nkcr, line 464: Not used.. yet?

cmd/mod.go
```diff
@@ -0,0 +1,100 @@
...
+	// Execute processes a command received from the CLI on the daemon.
+	Execute(Request) error
+}
+
+// Builder defines how to build a list of commands to tweak the ledger.
+type Builder interface {
+	// Command populates a new command.
+	Command(name string) CommandBuilder
+
+	Start(...Flag) Builder
```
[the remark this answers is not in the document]
Gilthoniel: Old habits on the build design pattern.

cmd/injector.go
```diff
@@ -0,0 +1,39 @@
...
+
+func (inj *reflectInjector) Resolve(v interface{}) error {
+	rv := reflect.ValueOf(v)
+	if rv.Kind() != reflect.Ptr {
+		return xerrors.New("expect a pointer")
+	}
+
+	for typ, value := range inj.mapper {
+		if typ.AssignableTo(rv.Elem().Type()) {
+			rv.Elem().Set(reflect.ValueOf(value))
```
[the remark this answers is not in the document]
Gilthoniel: The two approaches work I guess, I prefer this one because the work is done automatically.

mino/minogrpc/certs/mod.go
```diff
@@ -0,0 +1,41 @@
...
+
+	// Delete removes all the certificates associated with the address.
+	Delete(mino.Address)
+
+	// Range iterates over the certificates hold by the store.
+	Range(func(addr mino.Address, cert *tls.Certificate) bool)
+
+	// Fetch calls the address to fetch its certificate and verifies the
+	// integrity with the given digest.
+	Fetch(Dialable, []byte) error
```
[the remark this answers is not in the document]
Gilthoniel: Like any other error ? Why would you be more interested in the hash mismatch than a connection timeout for instance ?

mino/minogrpc/overlay.proto
Gilthoniel: It's a string. That's left to the implementation to generate bytes and convert them in hex (or base64 like it's doing now)

mino/minogrpc/tokens/mod.go
Gilthoniel: Not sure where you can eavesdrop as the request is sent over TLS.

mino/minogrpc/server.go
Gilthoniel: my mistake

Gilthoniel:
@nkcr Thanks a lot, I wasn't super happy with the abstraction but it now looks much better. I changed one thing from our discussion this morning which is the _Action_ for a node builder that is now a _ActionTemplate_ that you transformed into an actual action which allows to have the abstraction at the top level.

mino/minogrpc/certs/mod.go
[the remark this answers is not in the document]
nkcr: will be addressed by #55

PR #57, opened 2020-06-02
Draft: serde

Gilthoniel:
This is the first proposal for a encoding module. It focuses on a pipeline from the data model to a network message to finally a slice of bytes and reversely. A factory provides the primitives for the first step and the encoder for the second step.


Gilthoniel:
I'll be working on the follow up PRs to replace the encoding package. So for the moment, there's only the serde package.

serde/default.go
```diff
@@ -0,0 +1,51 @@
...
+
+const (
+	notImplementedErr = "not implemented"
+)
+
+// UnimplementedMessage is a default implementation of the message interface
+// which returns error to each function. It should be embedded to any message
+// implementation.
+//
+// - implement serde.Message
```
nkcr, line 13: until now we used `- implements` with a "s"

PR #62, opened 2020-06-15
Serde migration: clean up + qsc + dkg

Gilthoniel:
This is the last step before migrating mino. It removes encoding mentions in all packages except mino obviously.


consensus/qsc/messages.go
```diff
@@ -0,0 +1,181 @@
...
+// VisitJSON implements serde.Message.
+func (mset MessageSet) VisitJSON(ser serde.Serializer) (interface{}, error) {
+	messages := make(map[int64]json.Message)
+	for index, msg := range mset.messages {
+		raw, err := msg.VisitJSON(ser)
+		if err != nil {
+			return nil, xerrors.Errorf("couldn't serialize message: %v", err)
+		}
+
+		messages[index] = raw.(json.Message)
```
nkcr, line 88: this always makes me anxious, but yes, there is a priori no reason that `msg.VisiJSON` returns another type

Gilthoniel: Yes, it will panic in unit tests if someone changes it.

PR #63, opened 2020-06-16
Updates the DKG actor model

nkcr:


dkg/mod.go
```diff
@@ -7,12 +7,20 @@ import (
...
 // Actor defines the primitives to use a DKG protocol
 type Actor interface {
+	// Setup must be first called by ONE of the actor to use the subsequent
+	// functions. It creates the public distributed key and the private share on
+	// each node. Each node represented by a player must first execute Listen().
+	Setup(players mino.Players, pubKeys []kyber.Point, threshold int) error
+
 	Encrypt(message []byte) (K, C kyber.Point, remainder []byte, err error)
-	Decrypt(K, C kyber.Point) ([]byte, error)
+	Decrypt(players mino.Players, K, C kyber.Point) ([]byte, error)
```
Gilthoniel, line 23: Isn't the players included in the start result ?

dkg/pedersen/mod.go
```diff
@@ -22,66 +22,85 @@ var suite = suites.MustFind("Ed25519")
...
+func (a *Actor) Setup(players mino.Players, pubKeys []kyber.Point, threshold int) error {
+
 	if players.Len() != len(pubKeys) {
-		return nil, xerrors.Errorf("there should be as many players as "+
-			"pubKey: %d := %d", players.Len, len(pubKeys))
+		return xerrors.Errorf("there should be as many players as "+
+			"pubKey: %d := %d", players.Len(), len(pubKeys))
+	}
+
+	if a.startRes.distrKey != nil {
```
Gilthoniel, line 76: what if startRes is null ?

Gilthoniel: I would expect the _startRes_ pointer to be nil until a setup is done. It's weird that it is set but with missing data.

dkg/mod.go
nkcr: No, and the result is only seen by the leader. Since the nodes receive a `[]Address` in the start message, we could build from that a `mino.Players` so that each node has its players. But that seems a bit hacky.

dkg/pedersen/mod.go
nkcr: That's what I did first, but since the actor references it, if its nil when it references it, it won't be updated https://play.golang.org/p/1b-ignxqpGP
Thus, this is why I must initialize it first.

dkg/mod.go
Gilthoniel: What I mean is that a setup has a very well defined set of players so you shouldn't need to provide that information to decode.

dkg/pedersen/mod.go
Gilthoniel: Oh I see, I miss that. That's actually wrong because you only lock on the handler side but not on the actor one. That's why I was proposing a independent structure because it's much easier to manage and test. That's kinda what you did with the startRes thing but it's half-way done. In that case, it's having something like the following:
```go
type state struct {...}

func (s state) Done() bool { return s.isDone }

func (s state) GetDistKey() kyber.Point { return s.key }
```

Gilthoniel: with the locking and stuff

dkg/mod.go
nkcr: yes, but the setup is done once on the leader, so **only the leader** has access to this very well defined set of players. The other nodes have access to a slice of addresses, but we need a mino.Players to do the decrypt. 
If we want every actor to have the mino.Players, we could change `listen()` to take the mino.Players instead of taking it from `Setup()`.

Gilthoniel: Could this be useful: `mino.NewAddresses(addrs...)` ?

nkcr: that's perfect 😃

nkcr: I don't know if its a bug or not, but I was surprised that
```go
	for players.AddressIterator().HasNext() {
		addrs = append(addrs, players.AddressIterator().GetNext())
	}
```
wouldn't work and enter an infinite loop with the players returned by `mino.NewAddresses(addrs...)`. This is due to `AddressIterator()` returning each time a new fresh iterator.

Gilthoniel: Yeah you're not supposed to create a new iterator.

dkg/pedersen/handler.go
```diff
@@ -285,6 +314,11 @@ func (h *Handler) start(start Start, receivedDeals []Deal, from mino.Address,
 	h.privShare = distrKey.PriShare()
 	h.Unlock()
 
+	h.startRes.Lock()
+	h.startRes.distrKey = distrKey.Public()
+	h.startRes.participants = start.addresses
+	h.startRes.Unlock()
```
Gilthoniel, line 320: Can you move that to a function to be consistent ? In case the state is moved to a different package for instance.

mino/addr.go
```diff
@@ -22,12 +22,9 @@ func (it *addressIterator) HasNext() bool {
...
 func (it *addressIterator) GetNext() Address {
-	if it.HasNext() {
-		res := it.addrs[it.index]
-		it.index++
-		return res
-	}
-	return nil
+	res := it.addrs[it.index]
+	it.index++
+	return res
```
Gilthoniel, line 27: Thanks !

nkcr:
Had to update a test to make it pass on Mac OS

PR #67, opened 2020-06-30
Adds a first naive lottery package using DARCs

nkcr:


dkg/mod.go
```diff
@@ -17,7 +17,7 @@ type Actor interface {
 	// Setup must be first called by ONE of the actor to use the subsequent
 	// functions. It creates the public distributed key and the private share on
 	// each node. Each node represented by a player must first execute Listen().
-	Setup(players mino.Players, pubKeys []kyber.Point, threshold int) error
+	Setup(players mino.Players, pubKeys []kyber.Point, threshold int) (pubKey kyber.Point, err error)
```
Gilthoniel, line 20: Can you implement a `crypto.CollectiveAuthority` that wraps the addresses and the public keys ? You can implement the crypto ed25519 abstraction in the same time.

go.mod
```diff
@@ -3,16 +3,22 @@ module go.dedis.ch/dela
 go 1.13
 
 require (
-	github.com/golang/protobuf v1.3.4
+	github.com/cpuguy83/go-md2man/v2 v2.0.0 // indirect
```
Gilthoniel, line 6: I think those indirect imports will disappear with a go mod tidy

lottery/calypso/mod.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+// Setup implements lottery.Secret
+func (c *Calypso) Setup(players mino.Players, pubKeys []kyber.Point,
+	threshold int) (pubKey kyber.Point, err error) {
+
+	actor, err := c.dkg.Listen()
+	if err != nil {
+		return nil, xerrors.Errorf("failed to listen dkg: %v", err)
+	}
```
Gilthoniel, line 61: How do you handle multiple nodes ? All of them should use Listen but only one needs to do a setup.

lottery/calypso/mod.go
```diff
@@ -0,0 +1,238 @@
...
+		C:       cBuf,
+		DarcKey: darcKey,
+	}
+
+	messageBuf, err := json.Marshal(messageJSON)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to marshal message: %v", err)
+	}
+
+	c.storage.Store(key, messageBuf)
```
Gilthoniel, line 113: Why don't you use the hash to detect duplicates ? Also the access control should be stored alongside with the encrypted message.

lottery/mod.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"go.dedis.ch/dela/ledger/arc"
+	"go.dedis.ch/kyber/v3"
+)
+
+// Secret defines the primitives to run a Calypso-like app. In the case of
+// Calypso it is mainly a wrapper arround DKG that provides a storage and
+// authorization layer.
+type Secret interface {
```
Gilthoniel, line 11: I'm not sure to understand why those two interfaces are part of the lottery package as it is not specific to a lottery. You can simply have a calypso package with a _Calypso_ interface and an _EncryptedMessage_ interface. Let's keep applications out of the repo (or maybe in a integration test).

Gilthoniel: You can refer to the diagram and update it if necessary.

lottery/storage/mod.go
```diff
@@ -0,0 +1,7 @@
+package storage
+
+// KeyValue defines a simple key value storage
+type KeyValue interface {
+	Store(key, value []byte) error
```
Gilthoniel, line 5: You can use _serde.Message_ instead so that in-mem implementations don't need to serialize.

go.mod
nkcr: It's actually the `go mod tidy` that made them appear 🤔

lottery/calypso/mod.go
nkcr: That's a valid concern indeed. Could be solved by adding an argument `isLeader bool` and leave the responsibility to the user to only call Setup once with isLeader = true.
Then we should update the dkg.Actor to add a `GetPublicKey()` function, because right now the only way to get get the public key is by calling `actor.Setup()`.

lottery/calypso/mod.go
nkcr: you mean using `H(K || C)` for the key instead of using a random one?

lottery/calypso/mod.go
Gilthoniel: Yes with `GetPublicKey()` but the calypso interface should provide a `Listen` and a `Setup` separated.

lottery/mod.go
nkcr: Yep, lottery is not the best name. I did this to avoid having a `calypso.Calypso` interface, which isn't recommended.

lottery/calypso/mod.go
Gilthoniel: Yes, and now that I think of it, you can't update the access control if we don't have a ledger.

lottery/mod.go
Gilthoniel: To be fair, `calypso.Calypso` does not get any complain from go lint and it's much clearer than `lottery.Secret`.

calypso/mod.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"go.dedis.ch/dela/crypto"
+	"go.dedis.ch/dela/ledger/arc"
+	"go.dedis.ch/kyber/v3"
+)
+
+// Calypso defines the primitives to run a Calypso app. It is mainly a wrapper
+// arround DKG that provides a storage and authorization layer.
+type Calypso interface {
```
Gilthoniel, line 11: I guess you can call that `PrivateStorage` as Calypso is one possible implementation of a private storage.

calypso/calypso.go
```diff
@@ -0,0 +1,264 @@
...
+	return &Caly{
+		dkg:     dkg,
+		storage: inmemory.NewInMemory(),
+	}
+}
+
+// Caly is a wrapper around DKG to provide a storage and authorization layer
+//
+// implements calypso.Secret
+type Caly struct {
```
Gilthoniel, line 43: comment update + make it private

calypso/calypso.go
```diff
@@ -0,0 +1,264 @@
...
+	if c.dkgActor != nil {
+		return xerrors.Errorf("listen has already been called")
+	}
+
+	actor, err := c.dkg.Listen()
+	if err != nil {
+		return xerrors.Errorf("failed to listen dkg: %v", err)
+	}
+
+	c.dkgActor = actor
```
Gilthoniel, line 60: Sorry I just noticed that you can make your life easier by asking a DKG actor only. The one instantiating will take care of calling the `Listen` function. That's part of the idea of splitting the setup and actor logic.

Gilthoniel: Then you can remove the point to the structure.

calypso/calypso.go
```diff
@@ -0,0 +1,264 @@
...
+	if err != nil {
+		return nil, xerrors.Errorf("failed to marshal K: %v", err)
+	}
+
+	_, err = em.GetC().MarshalTo(&buf)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to marshal C: %v", err)
+	}
+
+	key := crypto.NewSha256Factory().New().Sum(buf.Bytes())
```
Gilthoniel, line 111: `Sum` does not do what you expect from it, for instance:
```go
func (d *digest) Sum(in []byte) []byte {
	// Make a copy of d so that caller can keep writing and summing.
	d0 := *d
	hash := d0.checkSum()
	if d0.is224 {
		return append(in, hash[:Size224]...)
	}
	return append(in, hash[:]...)

}
```

calypso/storage/inmemory/mod.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"go.dedis.ch/dela/serde"
+	serdej "go.dedis.ch/dela/serde/json"
+	"golang.org/x/xerrors"
+)
+
+// NewInMemory returns a new in memory database
+func NewInMemory() *InMemory {
+	return &InMemory{
+		database:   make(map[string][]byte),
```
Gilthoniel, line 12: Skip the serialization and store the data directly.

dkg/pedersen/mod.go
```diff
@@ -65,41 +68,50 @@ type Actor struct {
...
 	if err != nil {
-		return xerrors.Errorf("failed to stream: %v", err)
+		return nil, xerrors.Errorf("failed to stream: %v", err)
 	}
 
-	players.AddressIterator().Seek(0)
-	addrs := make([]mino.Address, 0, players.Len())
-	for players.AddressIterator().HasNext() {
-		addrs = append(addrs, players.AddressIterator().GetNext())
+	co.AddressIterator().Seek(0)
```
Gilthoniel, line 85: You don't need to seek an iterator that you're not going to use. The function *creates* an iterator.

dkg/pedersen/mod.go
```diff
@@ -65,41 +68,50 @@ type Actor struct {
...
+	addrs := make([]mino.Address, 0, co.Len())
+	pubkeys := make([]kyber.Point, 0, co.Len())
+	iter := co.AddressIterator()
+	for iter.HasNext() {
+		addr := iter.GetNext()
+		addrs = append(addrs, addr)
+		pubkey, index := co.GetPublicKey(addr)
```
Gilthoniel, line 92: Use two iterators: `co.AddressIterator()` and `co.PublicKeyIterator()` so you can avoid the search. Then you can remove the `PubKeyNotFound` thing. The implementation of the collective authority is responsible for the iterators to be consistent.

Gilthoniel: Actually even better: why don't you have the collective authority as an attribute ? So you don't need to transform stuff into arrays (I mean eventually yes, but it's implemented in the serialization logic, not here).

calypso/calypso.go
nkcr: not sure what you mean by "asking a DKG actor only"..?

Gilthoniel: Calypso can use the actor only, so instead of having New(dkg), you can have New(actor).

dkg/pedersen/mod.go
nkcr: I need the list of addresses anyway to call sender.send(). Since I can't fully skip transforming stuff into arrays I don't think its worth changing the attribute (plus it would just move the problem since at the end I still need those slices).

Gilthoniel: Ah yes, that's something to think about for the mino interface.

PR #69, opened 2020-07-07
Mino: fix sync of responses and errors with Call

Gilthoniel:
This fix the potential issue that occurs when an error is returned and the response channel is closed. The change makes sure that every error is read before closing.

### DIFF
--- a/mino/minoch/rpc.go
+++ b/mino/minoch/rpc.go
@@ -30,37 +30,45 @@ type RPC struct {
 }
 
 // Call sends the message to all participants and gather their reply. The
-// context in the scope of channel communication as there is no blocking I/O.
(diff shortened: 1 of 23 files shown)


mino/minoch/rpc.go
```diff
@@ -30,37 +30,45 @@ type RPC struct {
...
 	wg.Add(players.Len())
+
 	iter := players.AddressIterator()
+
 	for iter.HasNext() {
 		peer, err := c.manager.get(iter.GetNext())
 		if err != nil {
-			errs <- xerrors.Errorf("couldn't find peer: %v", err)
-			continue
+			// Abort everything is a peer is missing.
```
nkcr, line 55:
```suggestion
			// Abort everything if a peer is missing.
```

PR #74, opened 2020-07-10
Adds a first simple CLI for Calypso

nkcr:


nkcr:
`calypso/app` is for the moment an empty shell that could later hold the client CLI. So for now no need to look at it.

nkcr:
We moved part of this code to a new repo. This PR has been split to #76 and https://github.com/dedis/dela-apps/pull/1.

PR #75, opened 2020-07-14
Ledger refactoring

Gilthoniel:
This adds a _core_ module where all the ledger related packages will be moved eventually. For this PR, it only implements very simple dummy packages for the interfaces, then the already implemented ones will be moved progressively.


core/ordering/mod.go
```diff
@@ -0,0 +1,43 @@
...
+// to order transactions from a pool.
+type Service interface {
+	// Listen opens the endpoints of the service so that other participants can
+	// contact the node.
+	Listen() error
+
+	// Close closes the service.
+	Close() error
+
+	// GetProof must return a proof of the value of the provided key.
```
nkcr, line 37: ...value AT the...

core/tap/pool/mem/mod_test.go
```diff
@@ -0,0 +1,112 @@
...
+func TestPool_Add(t *testing.T) {
+	pool := NewPool()
+
+	err := pool.Add(fakeTx{id: []byte{1}})
+	require.NoError(t, err)
+
+	err = pool.Add(fakeTx{id: []byte{2}})
+	require.NoError(t, err)
+
+	// A transaction that exists in the active queue can simply be overriden
```
nkcr, line 34:
```suggestion
	// A transaction that exists in the active queue can simply be overwritten
```

core/execution/baremetal/mod.go
```diff
@@ -0,0 +1,37 @@
...
+func NewExecution(exec execution.Service) BareMetal {
+	return BareMetal{
+		exec: exec,
+	}
+}
+
+// Execute implements execution.Service. It uses the executor to process the
+// incoming transaction and return the result.
+func (bm BareMetal) Execute(tx tap.Transaction, snap store.Snapshot) (execution.Result, error) {
+	res, err := bm.exec.Execute(tx, snap)
```
nkcr, line 31: an execution.Service that uses another execution.Service ? Looks like the snake that bites its own tail.

core/tap/anon/mod.go
```diff
@@ -0,0 +1,99 @@
...
+	return nil
+}
+
+// GetArg implements tap.Transaction. It returns the value of the argument if it
+// is set, otherwise nil.
+func (t Transaction) GetArg(key string) []byte {
+	return t.args[key]
+}
+
+// Fingerprint implements serde.Fingerprinter. It writes a deteministic binary
```
nkcr, line 87:
```suggestion
// Fingerprint implements serde.Fingerprinter. It writes a deterministic binary
```

core/ordering/mod.go
```diff
@@ -0,0 +1,43 @@
...
+
+// Service is the interface of an ordering service. It provides the primitives
+// to order transactions from a pool.
+type Service interface {
+	// Listen opens the endpoints of the service so that other participants can
+	// contact the node.
+	Listen() error
+
+	// Close closes the service.
+	Close() error
```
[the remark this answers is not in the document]
Gilthoniel: Not sure if that will be useful but right you can do it yes, so I'll rename.

core/execution/baremetal/mod.go
Gilthoniel: Yeah it looks dump right now, but the goal was to register different services (darc, roster, etc...) that are the bare-metal smart contracts and this service takes care of looking up the right one.

Gilthoniel: Which will come while migrating the other packages.

PR #76, opened 2020-07-14
Moves Calypso and adds httpclient

nkcr:
Supersede #74


cli/node/injector.go
```diff
@@ -29,6 +29,10 @@ func (inj *reflectInjector) Resolve(v interface{}) error {
 		return xerrors.New("expect a pointer")
 	}
 
+	if !rv.Elem().IsValid() {
+		return xerrors.Errorf("reflect value '%v' is invalid", rv)
```
Gilthoniel, line 33: Can you complete the unit test for that statement? I think the `%v` might be too verbose.

dkg/pedersen/controller/mod.go
```diff
@@ -0,0 +1,55 @@
...
+	var no mino.Mino
+	err := inj.Resolve(&no)
+	if err != nil {
+		return xerrors.Errorf("failed to resolve mino: %v", err)
+	}
+
+	privkey := dkg.Suite.Scalar().Pick(dkg.Suite.RandomStream())
+	pubkey := dkg.Suite.Point().Mul(privkey, nil)
+
+	dkg := pedersen.NewPedersen(privkey, no)
```
Gilthoniel, line 41: Instead of asking for the private key, you should let the constructor generates it (and later we can improve to load an existing one from a file or whatever). Then you can remove the public _Suite_ variable.

dkg/mod.go
```diff
@@ -3,8 +3,12 @@ package dkg
 import (
 	"go.dedis.ch/dela/crypto"
 	"go.dedis.ch/kyber/v3"
+	"go.dedis.ch/kyber/v3/suites"
 )
 
+// Suite is the Kyber suite for Pedersen.
+var Suite = suites.MustFind("Ed25519")
```
Gilthoniel, line 10: Anyone can change it so we should do it differently. See the previous comment.

dkg/pedersen/handler.go
```diff
@@ -118,10 +119,10 @@ mainSwitch:
 
 		// TODO: check if started before
 		h.RLock()
-		S := suite.Point().Mul(h.privShare.V, msg.K)
+		S := dkg.Suite.Point().Mul(h.privShare.V, msg.K)
```
Gilthoniel, line 122: Keep using the private field one.

dkg/pedersen/mod.go
```diff
@@ -10,15 +10,11 @@ import (
...
 	"go.dedis.ch/kyber/v3"
 	"go.dedis.ch/kyber/v3/share"
-	"go.dedis.ch/kyber/v3/suites"
 	"go.dedis.ch/kyber/v3/util/random"
 	"golang.org/x/net/context"
 	"golang.org/x/xerrors"
 )
 
-// Suite is the Kyber suite for Pedersen.
-var suite = suites.MustFind("Ed25519")
```
Gilthoniel, line 20: See other comments.

ledger/byzcoin/controller/mod.go
```diff
@@ -35,9 +37,10 @@ func (m minimal) SetCommands(builder node.Builder) {
 // Run implements node.Initializer.
 func (m minimal) Inject(ctx cli.Flags, inj node.Injector) error {
 	var no mino.Mino
+	fmt.Println("asking for:", &no)
```
Gilthoniel, line 40: debugging ?

mino/httpclient/mod.go
```diff
@@ -0,0 +1,140 @@
+package httpclient
```
Gilthoniel, line 1: It looks to me it is doing the job of a proxy, so what do you think about `proxy` ? Then you have the HTTP proxy implementation.

mino/httpclient/controller/mod.go
```diff
@@ -0,0 +1,45 @@
+package controller
```
Gilthoniel, line 1: tests

mino/httpclient/mod.go
Gilthoniel: also missing tests

mino/httpclient/mod.go
```diff
@@ -0,0 +1,140 @@
...
+func (s Client) Start() {
+	s.logger.Println("Client server is starting...")
+
+	done := make(chan bool)
+	quit := make(chan os.Signal, 1)
+	signal.Notify(quit, os.Interrupt)
+
+	go func() {
+		<-quit
+		s.logger.Println("Server is shutting down...")
```
Gilthoniel, line 72: You can make it simpler (and easier to test) by removing the signal catching and having _Start_ and _Stop_ functions. When the daemon is stopped, it should be possible to call _Stop_ but it is yet to be implemented.

Gilthoniel: Also rename Start to Listen and the function can be blocking.

mino/httpclient/mod.go
```diff
@@ -0,0 +1,140 @@
...
+	// Start starts the http server. This call is assumed to be blocking
+	Start()
+
+	// RegisterHandler registers a new handler
+	RegisterHandler(path string, handler func(http.ResponseWriter, *http.Request)) error
+}
+
+// NewClient creates a new http client
+func NewClient(listenAddr string) *Client {
+	logger := log.New(os.Stdout, "httpclient :", log.LstdFlags)
```
Gilthoniel, line 33: `logger := log.New(dela.Logger, "", 0)` to keep consistent logging.

Gilthoniel: https://github.com/rs/zerolog#integration-with-nethttp for out of the box support.

dkg/pedersen/controller/mod.go
nkcr: mmm thing is the use need the associated public key. So are you suggesting the constructor also return the generated public key ?

Gilthoniel:
@nkcr check the failing test, there's a reuse of ports.

nkcr:
Indeed, tests are so fast that the server wasn't yet fully stopped before the start of the next test. Fixed!

PR #77, opened 2020-07-17
Merkle tree

Gilthoniel:
This PR refactors the store to be generic and adds a specialization to provide proof of inclusion and absence through hash trees. It also adds the Merkle binary prefix tree implementation.


core/store/hashtree/mem/tree.go
```diff
@@ -0,0 +1,477 @@
...
+
+	Prepare(nonce []byte, prefix *big.Int, fac crypto.HashFactory) ([]byte, error)
+
+	Visit(func(TreeNode))
+
+	Clone() TreeNode
+}
+
+// Tree is an implementation of a Merkle binary prefix tree. Due to the
+// structure of the tree, any prefix of an index is overriden which means that
```
nkcr, line 51:
```suggestion
// structure of the tree, any prefix of an index is overridden which means that
```

nkcr: \+ same as previous, not sure what index you refer to.

core/store/hashtree/mem/tree.go
```diff
@@ -0,0 +1,477 @@
...
+	t.root.Visit(func(n TreeNode) {
+		if n.GetType() == leafNodeType {
+			counter++
+		}
+	})
+
+	return counter
+}
+
+// Search returns the value associated to the key if it exists, otherwise nil.
```
nkcr, line 84: I would comment the role of the path argument

core/store/hashtree/mem/tree.go
```diff
@@ -0,0 +1,477 @@
...
+// Nonce is the type of the tree nonce.
+type Nonce [8]byte
+
+const (
+	// DepthLength is the length in bytes of the binary representation of the
+	// depth.
+	DepthLength = 2
+
+	// MaxDepth is the maximum depth the tree should reach. It is equivalent to
+	// the maximum key length.
```
nkcr, line 20:
```suggestion
	// the maximum key length in bytes.
```

core/store/hashtree/mem/tree.go
```diff
@@ -0,0 +1,477 @@
...
+}
+
+// Delete implements mem.TreeNode. It ignores the delete as an empty node
+// already means the key is missing.
+func (n *EmptyNode) Delete(key *big.Int) TreeNode {
+	return n
+}
+
+// Prepare implements mem.TreeNode. It updates the hash of the node and return
+// the digest.
```
nkcr, line 200:
```suggestion
// Prepare implements mem.TreeNode. It updates the hash of the node and return
// the digest.
//
// We build and hash the following datagram:
//
// | TYPE | NONCE | PREFIX | DEPTH |
```

core/store/hashtree/mem/path.go
```diff
@@ -0,0 +1,70 @@
...
+
+// Path is a path along the tree nodes to a key and its value, or none.
+//
+// - implements hashtree.Path
+type Path struct {
+	key []byte
+	// Root is the root of the hash tree. This value is not serialized and
+	// reproduced from the leaf and the interior nodes when deserializing.
+	root      []byte
+	interiors [][]byte
```
nkcr, line 17: it should be clear that it's only the hashes of the interior nodes, and not their keys. Which make me wonder, why not storing the keys instead of the hashes? Given a tree, if one gives me the path from a child to the node, I am expecting getting all the interior nodes, not only a list of hashes.

core/store/hashtree/mem/tree.go
```diff
@@ -0,0 +1,477 @@
...
+
+	if path != nil {
+		path.interiors = append(path.interiors, n.left.GetHash())
+	}
+
+	return n.right.Search(key, path)
+}
+
+// Insert implements mem.TreeNode. It inserts the key/value pair to the right
+// path.
```
nkcr, line 279:
```suggestion
// Insert implements mem.TreeNode. It inserts the key/value pair to the right or left based on the provided bi path.
```

core/store/hashtree/mem/tree.go
```diff
@@ -0,0 +1,477 @@
...
+	if bi.Bit(int(n.depth)) == 0 {
+		n.left = n.left.Insert(key, value, bi)
+	} else {
+		n.right = n.right.Insert(key, value, bi)
+	}
+
+	return n
+}
+
+// Delete implements mem.TreeNode. It deletes the key from the right path.
```
nkcr, line 290:
```suggestion
// Delete implements mem.TreeNode. It deletes the key from the right of left path, based on the bi path.
```

The naming is fragile here, in this function we have a `key` which is a `*big.Int`, but in the previous `Insert()` one we have a `key` that is a `[]byte` and then another `*big.Int` argument which has the same utility as here but is called there `bi`.

nkcr: This is a global remark: it can be hard to get around all the different notions of `key` `index` `prefix` `path` `value` `depth` that sometimes use `[]byte` or `*big.Int`, or is even converted to `int`.

I would find really useful to have a small example of the tree with what and how things are stored. Sometimes just having a small representation does 90% of the job (a bit like the routing tree: https://github.com/dedis/dela/blob/master/mino/minogrpc/routing/routing.go#L282)

core/store/hashtree/mem/tree.go
```diff
@@ -0,0 +1,477 @@
...
+	if n.left.GetType() == emptyNodeType && n.right.GetType() == emptyNodeType {
+		// If an interior node points to two empty nodes, it is itself an empty
+		// one.
+		return NewEmptyNode(n.depth)
+	}
+
+	return n
+}
+
+// Prepare implements mem.TreeNode. It updates the hash of the node and return
```
nkcr, line 307:
```suggestion
// Prepare implements mem.TreeNode. It updates the hash of the node and returns
```

core/store/hashtree/mem/tree.go
```diff
@@ -0,0 +1,477 @@
...
+// the digest.
+func (n *InteriorNode) Prepare(nonce []byte, prefix *big.Int, fac crypto.HashFactory) ([]byte, error) {
```
nkcr, line 309: hard width limit reached ; )

core/store/hashtree/mem/tree.go
```diff
@@ -0,0 +1,477 @@
...
+
+	node := NewInteriorNode(n.depth)
+
+	if curr.Bit(int(n.depth)) == 0 {
+		node.left = node.left.Insert(n.key, n.value, curr)
+	} else {
+		node.right = node.right.Insert(n.key, n.value, curr)
+	}
+
+	if bi.Bit(int(n.depth)) == 0 {
```
nkcr, line 409: what if both `if` are true? Is it always mutually exclusive ?

core/store/hashtree/mem/mod.go
```diff
@@ -0,0 +1,119 @@
...
+
+	"go.dedis.ch/dela/core/store"
+	"go.dedis.ch/dela/core/store/hashtree"
+	"go.dedis.ch/dela/crypto"
+	"golang.org/x/xerrors"
+)
+
+// MerkleTree is an in-memory implementation of a Merkle prefix binary tree.
+// This particular implementation assumes the keys will have the same length so
+// that any prefix is overriden by the index.
```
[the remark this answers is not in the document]
Gilthoniel: They use _index_ in the paper thus the usage here but it's simply the key.

core/store/hashtree/mem/mod.go
```diff
@@ -0,0 +1,119 @@
...
+func (t *MerkleTree) GetRoot() []byte {
+	return t.tree.root.GetHash()
+}
+
+// GetPath implements hashtree.Tree. It returns a path to a given key that can
+// be used to prove the inclusion or the absence of a key.
+func (t *MerkleTree) GetPath(key []byte) (hashtree.Path, error) {
+	path := newPath(key)
+
+	_, err := t.tree.Search(key, &path)
```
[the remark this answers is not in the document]
Gilthoniel: Yes you want to search with or without the proof.

core/store/hashtree/mem/mod.go
```diff
@@ -0,0 +1,119 @@
...
+	if err != nil {
+		return nil, xerrors.Errorf("couldn't search key: %v", err)
+	}
+
+	return path, nil
+}
+
+// Stage implements hashtree.Tree. It executes the callback over a child of the
```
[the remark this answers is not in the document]
Gilthoniel: Yes, over a clone.

core/store/hashtree/mem/path.go
Gilthoniel: The key is not useful and you expect the proof to contain the minimum set of information that will allow you to check the integrity.

core/store/hashtree/mem/tree.go
Gilthoniel: In all cases, it's still the key but it can actually be always a *big.Int for simplicity. For the example, the implementation is based on a paper that explains everything so I'm not really keen to write a different (and probably bad) explanation on how it works where the paper is really smooth in that regard.

core/store/hashtree/mem/tree.go
Gilthoniel: Yes it is

nkcr: ok, so maybe we could merge those two ifs? Or maybe simply a comment?

core/store/hashtree/mem/path.go
```diff
@@ -0,0 +1,70 @@
...
+func (s Path) GetValue() []byte {
+	switch leaf := s.leaf.(type) {
+	case *LeafNode:
+		return leaf.value
+	default:
+		return nil
+	}
+}
+
+// GetRoot implements hashtree.Path. It returns the root of the path.
```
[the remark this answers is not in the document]
Gilthoniel: The hash is not key actually. the key is an empty prefix.

core/store/hashtree/mem/tree.go
Gilthoniel: Similarly to the other comment, the paper explains in details how the hash is created, and *why*. A pointer to the paper should be sufficient ?

core/store/hashtree/mem/tree.go
Gilthoniel: You definitely can't merge them, it would be wrong. But a comment is feasible!

PR #84, opened 2020-08-12
Core: view change as a transaction

Gilthoniel:
This adds an execution service to change the roster through transactions.

### DIFF
--- a/core/execution/baremetal/viewchange/mod.go
+++ b/core/execution/baremetal/viewchange/mod.go
@@ -1,0 +1,83 @@
+package viewchange
+
+import (
+	"go.dedis.ch/dela/consensus/viewchange"
+	"go.dedis.ch/dela/core/execution"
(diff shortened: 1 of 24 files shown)


Gilthoniel:
@nkcr You can ignore the json package for now, the tests are missing.

core/execution/baremetal/viewchange/mod.go
```diff
@@ -0,0 +1,88 @@
...
+
+	currData, err := snap.Get(c.rosterKey)
+	if err != nil {
+		res.Message = messageStorageEmpty
+		return res, xerrors.Errorf("failed to read roster: %v", err)
+	}
+
+	curr, err := c.rosterFac.AuthorityOf(c.context, currData)
+	if err != nil {
+		res.Message = messageStorageEmpty
```
nkcr, line 65: this error message is not really accurate, plus having two times the same one could make things hard to debug.

PR #86, opened 2020-08-13
Make transaction pool easier to use

Gilthoniel:
Needs #84

### DIFF
--- a/core/txn/pool/gatherer.go
+++ b/core/txn/pool/gatherer.go
@@ -1,0 +1,76 @@
+package pool
+
+import (
+	"context"
+	"sync"
+
+	"go.dedis.ch/dela/core/txn"
+)
+
+// Gatherer is a common tool to the pool implementations that helps to implement
+// the gathering process.
+type Gatherer interface {
+	// Wait waits for a notification with sufficient transactions to return the
+	// array, or nil if the context ends.
+	Wait(ctx context.Context, cfg Config) []txn.Transaction
+
+	// Notify triggers any wait that fits the given lenght. The function is only
+	// called when necessary.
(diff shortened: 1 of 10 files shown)


core/txn/pool/mod.go
```diff
@@ -10,20 +10,21 @@ import (
...
 )
 
-// Event is an event triggered when new transactions arrived to the pool.
-type Event struct {
-	// Len is the current length of the pool.
-	Len int
+// Config is the set of parameters that allows one to change the behavior of the
+// gathering process.
+type Config struct {
+	// Min indicates what is minimum number of transactions that is required
```
nkcr, line 16:
```suggestion
	// Min indicates what is the minimum number of transactions required
```

core/txn/pool/gatherer.go
```diff
@@ -0,0 +1,76 @@
...
+// Gatherer is a common tool to the pool implementations that helps to implement
+// the gathering process.
+type Gatherer interface {
+	// Wait waits for a notification with sufficient transactions to return the
+	// array, or nil if the context ends.
+	Wait(ctx context.Context, cfg Config) []txn.Transaction
+
+	// Notify triggers any wait that fits the given lenght. The function is only
+	// called when necessary.
+	Notify(length int, fn func() []txn.Transaction)
```
nkcr, line 19: are you using a callback for performance reasons? If so I would explain why in the comment. I guess it's a good reason, but it make the code painful to read and understand.

Gilthoniel: Yes, I didn't want to make an array every time a transaction is added. But actually I had to change the gatherer in the next PR because of the locks (txs could be ignored) and the notify will disappear anyway.

PR #88, opened 2020-08-14
Core: use db transaction when required

Gilthoniel:
Needs #87

### DIFF
--- a/core/store/kv/default.go
+++ b/core/store/kv/default.go
@@ -30,27 +30,17 @@ func New(path string) (DB, error) {
 
 // View implements kv.DB. It opens a read-only transaction and opens the
 // provided bucket. It will return an error if the bucket does not exist.
-func (db boltDB) View(bucket []byte, fn func(Bucket) error) error {
+func (db boltDB) View(fn func(ReadableTx) error) error {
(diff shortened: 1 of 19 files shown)


core/store/kv/default.go
```diff
@@ -30,27 +30,17 @@ func New(path string) (DB, error) {
...
-			return xerrors.Errorf("bucket '%x' not found", bucket)
-		}
-
-		return fn(boltBucket{bucket: b})
+		return fn(boltTx{txn: txn})
 	})
 }
 
 // Update implements kv.DB. It opens a read-write transaction and opens the
 // bucket. It will create it if it does not exist yet.
```
nkcr, line 40: comment update

core/store/mod.go
```diff
@@ -18,3 +18,9 @@ type Snapshot interface {
 	Readable
 	Writable
 }
+
+// Transaction is a generic interface that store implementations can use to
+// provide atomicity.
```
nkcr, line 23: I would also say that it "adds a handler function to be executed after the transaction successfully commits."

nkcr (approved):

PR #91, opened 2020-09-01
Implement the proof interface

Gilthoniel:
This implements the proof interface for cosipbft and it also uses the chain in the synchronization process.

### DIFF
--- a/core/ordering/cosipbft/blocksync/default_test.go
+++ b/core/ordering/cosipbft/blocksync/default_test.go
@@ -206,8 +206,8 @@ func TestHandler_Stream(t *testing.T) {
 	require.EqualError(t, err, "sending ack failed: fake error")
 }
 
-// Utility functions
 // -----------------------------------------------------------------------------
+// Utility functions
 
 func makeChain(t *testing.T, index uint64) otypes.Chain {
 	block, err := otypes.NewBlock(simple.NewData(nil), otypes.WithIndex(index))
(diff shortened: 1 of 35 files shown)


Gilthoniel:
The json format cleaning and tests will come in a following PR.

core/ordering/cosipbft/blocksync/default_test.go
```diff
@@ -169,46 +181,71 @@ func TestHandler_Stream(t *testing.T) {
...
 	err = handler.Stream(fake.Sender{}, fake.NewReceiver(msgs...))
 	require.Error(t, err)
 	require.Regexp(t, "pbft catch up failed: mismatch link '[0]{8}' != '[0-9a-f]{8}'", err.Error())
 
-	err = handler.Stream(fake.NewBadSender(), fake.NewReceiver(types.NewSyncMessage(0)))
+	err = handler.Stream(fake.NewBadSender(), fake.NewReceiver(types.NewSyncMessage(makeChain(t, 0))))
 	require.EqualError(t, err, "sending ack failed: fake error")
 }
 
 // Utility functions -----------------------------------------------------------
```
nkcr, line 209: I don't mind changing to this, but then we should stay consistent with the rest, as we normally use

```
// -----------------------------------------------------------------------------
// Utility functions
```

Gilthoniel: I replaced the occurrences.

nkcr: you put the line at the bottom, instead of top 😅

Gilthoniel: oopsy !

nkcr (approved):

PR #93, opened 2020-09-02
Remove unused packages from the refactoring

Gilthoniel:
This removes the unused packages while keeping the ones with dependencies. They will be moved in a different PR.


nkcr (approved):
Good bye, old packages. You have served us well... 👋😢

PR #100, opened 2020-09-08
Core: pbft view are now signed

Gilthoniel:
This adds a signature to view messages so that it can be verified that a node asked for a view change.

Also: add view messages when sending a proposed block if the leader comes from a view change so that nodes can catch up if necessary.

### DIFF
--- a/core/ordering/cosipbft/json/mod.go
+++ b/core/ordering/cosipbft/json/mod.go
@@ -310,7 +310,7 @@ func (f msgFormat) Encode(ctx serde.Context, msg serde.Message) ([]byte, error) 
 	case types.ViewMessage:
 		vm, err := encodeView(in, ctx)
 		if err != nil {
-			return nil, err
+			return nil, xerrors.Errorf("view: %v", err)
 		}
 
 		m = MessageJSON{View: vm}
--- a/core/ordering/cosipbft/json/mod.go
+++ b/core/ordering/cosipbft/json/mod.go
@@ -7,6 +7,7 @@ import (
 	"go.dedis.ch/dela/core/ordering/cosipbft/types"
 	"go.dedis.ch/dela/core/validation"
 	"go.dedis.ch/dela/crypto"
+	"go.dedis.ch/dela/mino"
 	"go.dedis.ch/dela/serde"
 	"golang.org/x/xerrors"
 )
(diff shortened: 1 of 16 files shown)


core/ordering/cosipbft/json/mod.go
```diff
@@ -289,12 +308,12 @@ func (f msgFormat) Encode(ctx serde.Context, msg serde.Message) ([]byte, error)
 
 		m = MessageJSON{Done: &dm}
 	case types.ViewMessage:
-		vm := ViewMessageJSON{
-			ID:     in.GetID().Bytes(),
-			Leader: in.GetLeader(),
+		vm, err := encodeView(in, ctx)
+		if err != nil {
```
nkcr, line 312: error wrap

core/ordering/cosipbft/pbft/mod.go
```diff
@@ -254,67 +274,126 @@ func (m *pbftsm) Finalize(id types.Digest, sig crypto.Signature) error {
...
+	m.round.views[view.from] = view
 
 	m.checkViewChange(view)
+
+	return nil
 }
 
 // AcceptAll implements pbft.StateMachine. It accepts a list of views which
 // allows a node falling behind to catch up. The list must contain enough views
 // reach the threshold otherwise it will be ignored.
```
nkcr, line 323:
```suggestion
// to reach the threshold, otherwise it will be ignored.
```

core/ordering/cosipbft/pbft/mod.go
```diff
@@ -254,67 +274,126 @@ func (m *pbftsm) Finalize(id types.Digest, sig crypto.Signature) error {
...
 	defer m.Unlock()
 
-	set := make(map[mino.Address]struct{})
+	err := m.init()
+	if err != nil {
+		return xerrors.Errorf("init: %v", err)
+	}
+
+	if len(views) <= m.round.threshold {
+		return xerrors.New("not enough views")
```
nkcr, line 334:
```suggestion
		return xerrors.New("not enough views: %d <= %d", len(views), m.round.threshold)
```

core/ordering/cosipbft/pbft/view.go
```diff
@@ -0,0 +1,91 @@
...
+	err := pubkey.Verify(v.bytes(), v.signature)
+	if err != nil {
+		return xerrors.Errorf("verify: %v", err)
+	}
+
+	return nil
+}
+
+func (v View) bytes() []byte {
+	buffer := make([]byte, 8)
```
nkcr, line 87: 2 bytes should be enough

Gilthoniel: oops, yes I forgot to change it.

nkcr (approved):

PR #105, opened 2020-09-13
Minogrpc: memory management for the queue

Gilthoniel:


mino/minogrpc/session/queue.go
```diff
@@ -66,9 +85,29 @@ func (q *NonBlockingQueue) pushAndWait() {
...
 
 		// Wait for the channel to be available to writings.
 		q.ch <- msg
 	}
 }
+
+func (q *NonBlockingQueue) replaceBuffer() bool {
+	exp := float64(0)
+	if cap(q.buffer) > 0 {
+		exp = math.Floor(math.Log2(float64(cap(q.buffer)))) + 1
```
nkcr, line 99: I think you meant:
```suggestion
		exp = math.Floor(math.Log2(float64(cap(q.buffer)/q.cap))) + 1
```

Otherwise you go from a size of 100 to 128'000. There you go from 100 to 200.

nkcr: then you can do:
```
newSize := cap(q.buffer)*2
exp := math.Floor(math.Log2(float64(newSize))))
if exp > q.limit {
...
```

Gilthoniel: Geez, that's what happen when you choose badly the unit test values... of course with an initial capacity of 1, it's not going to detect that. Nice catch.

For the second comment, I can't do that because the capacity of the buffer evolves when taking a packet out:
```
buffer := make([]byte, 10) // cap == 10
buffer = buffer[1:] // cap == 9
```
so to keep the power of 2's sizes, I need to recalculate it.

nkcr (approved):

PR #108, opened 2020-09-16
Access: permissioned view change

Gilthoniel:


core/access/darc/json/mod.go
```diff
@@ -0,0 +1,153 @@
...
+	data, err := ctx.Marshal(m)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to marshal: %v", err)
+	}
+
+	return data, nil
+}
+
+func encodeExpression(ctx serde.Context, expr *types.Expression) (m ExpressionJSON, _ error) {
+	identities := make(types.IdentitySet, 0)
```
nkcr, line 63:
```suggestion
func encodeExpression(ctx serde.Context, expr *types.Expression) (ExpressionJSON, error) {
	var m EspressionJSON
	identities := make(types.IdentitySet, 0)
```

core/access/mod.go
```diff
@@ -3,33 +3,51 @@ package access
...
+//     "myContract:sayBye": ["Bob"],
+//   }
+//
+// The example above shows an access control for the contract "myContract" that
+// is allowing two commands "sayHello" and "sayBye". Alice and Bob can say
+// hello, but only Bob is allow to say bye. Alice can prove that she's allowed
+// by providing credentials with the identifier 0xdeadbeef.
+type Credentials interface {
+	GetID() []byte
+	GetRule() string
```
nkcr, line 40: from the example above I am missing what GetRules() corresponds to. Should this not return a `[]string` ?

core/access/darc/types/expr.go
```diff
@@ -0,0 +1,112 @@
...
+		matches: sets,
+	}
+}
+
+// GetIdentitySets returns the list of identity sets.
+func (expr *Expression) GetIdentitySets() []IdentitySet {
+	return append([]IdentitySet{}, expr.matches...)
+}
+
+// Evolve returns a new expression with the group added in the list of
```
nkcr, line 79: it actually doesn't return anything

core/access/darc/types/permission.go
```diff
@@ -0,0 +1,161 @@
...
+		if !grant {
+			return
+		}
+
+		expr = NewExpression()
+	}
+
+	expr.Evolve(grant, group)
+
+	if len(expr.matches) == 0 {
```
nkcr, line 82: this stuff is hard to get when reading the code. Why would it delete the rule in case the expression evolve doesn't have any match ?

core/access/darc/types/mod.go
```diff
@@ -0,0 +1,29 @@
...
+
+// Permission is the interface of the underlying permissions used by the
+// service.
+type Permission interface {
+	serde.Message
+
+	// Evolve grants or denies the permission to the rule to the group of
+	// identities as a single entity so that it will match if and only if the
+	// group agrees.
+	Evolve(rule string, grant bool, group ...access.Identity)
```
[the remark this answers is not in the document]
Gilthoniel: That's left to a future implementation ;). Basically you can either generate the different subgroups to allow a threshold (of course that's costly) or indeed add an acceptance threshold.

core/access/darc/json/mod.go
Gilthoniel: Espresso, what else !

core/access/mod.go
Gilthoniel: No because you match a single string, which can be whatever the credentials defines it should be. In the case of a contract, it's the contract and the action concatenated. But that could be a simple word for something else.

core/access/darc/types/permission.go
Gilthoniel: I guess the evolve should be split into Allow and Deny as this is just a simple cleaning procedure after removing a group.

Gilthoniel: well there was a bug with the removal so it is indeed better when split.

core/access/mod.go
nkcr: And in the case of the example above ?

Gilthoniel: Oh I see where the confusion is. The key on the left is the rule, not the array of identities on the right.

Gilthoniel: So in the example, the credentials are define with `0xdeadbeef`+`myContract:sayHello` or `0xdeadbeef` + `myContract:sayBye`

nkcr: Ok, so based on this example there would be 2 credential, right? Then maybe it would make more sense if `Credentials` is in singular ?

Gilthoniel: I had this thought too so probably, yeah.

nkcr: Ok cool. Then it would help if you can update the example to explain what GetRule corresponds to

Gilthoniel: The comment is also updated to avoid the confusion.

nkcr (approved):

PR #109, opened 2020-09-16
Disable logging and allow to set the level with an env variable

Gilthoniel:
Also: increase round timeout because of slow machines.


nkcr (approved):
Can be merged once the conflict is solved

PR #113, opened 2020-09-23
Core: persistent stores

Gilthoniel:
Also: change the block index logic to start at 0 instead of assuming the genesis block is the zero index.

### DIFF
--- a/core/store/hashtree/binprefix/mod.go
+++ b/core/store/hashtree/binprefix/mod.go
@@ -30,6 +30,8 @@
 package binprefix
 
 import (
+	"sync"
+
 	"go.dedis.ch/dela/core/store"
 	"go.dedis.ch/dela/core/store/hashtree"
 	"go.dedis.ch/dela/core/store/kv"
@@ -43,6 +45,8 @@ import (
 //
 // - implements hashtree.Tree
 type MerkleTree struct {
+	sync.Mutex
+
 	tree        *Tree
 	db          kv.DB
 	tx          store.Transaction
@@ -63,6 +67,9 @@ func NewMerkleTree(db kv.DB, nonce Nonce) *MerkleTree {
 // Load tries to read the bucket and scan it for existing leafs and populate the
 // tree with them.
 func (t *MerkleTree) Load() error {
+	t.Lock()
+	defer t.Unlock()
+
 	return t.doUpdate(func(tx kv.WritableTx) error {
 		bucket := tx.GetBucket(t.bucket)
 
@@ -83,6 +90,9 @@ func (t *MerkleTree) Load() error {
 // Get implements store.Readable. It returns the value associated with the key
 // if it exists, otherwise it returns nil.
 func (t *MerkleTree) Get(key []byte) ([]byte, error) {
+	t.Lock()
+	defer t.Unlock()
+
 	var value []byte
 
 	err := t.doView(func(tx kv.ReadableTx) error {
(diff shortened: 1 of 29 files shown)


core/store/hashtree/binprefix/mod.go
```diff
@@ -60,9 +64,35 @@ func NewMerkleTree(db kv.DB, nonce Nonce) *MerkleTree {
...
+
+	return t.doUpdate(func(tx kv.WritableTx) error {
+		bucket := tx.GetBucket(t.bucket)
+
+		err := t.tree.Load(bucket)
+		if err != nil {
+			return xerrors.Errorf("failed to load: %v", err)
+		}
+
+		err = t.tree.Update(t.hashFactory, bucket)
```
nkcr, line 81: I find  `t.tree.Load(bucket)` and `t.tree.Update(...)` to be very  meaningless. But I guess it's hard to make that code more explicit at this stage.

core/store/hashtree/binprefix/mod.go
```diff
@@ -60,9 +64,35 @@ func NewMerkleTree(db kv.DB, nonce Nonce) *MerkleTree {
 	}
 }
 
+// Load tries to read the bucket and scan it for existing leafs and populate the
```
nkcr, line 67: I'm glad there is the comment. I would've never understood that by reading the code!

core/store/hashtree/binprefix/tree.go
```diff
@@ -103,6 +103,73 @@ func (t *Tree) Len() int {
...
+	})
+
+	if err != nil {
+		return xerrors.Errorf("while scanning: %v", err)
+	}
+
+	return nil
+}
+
+func (t *Tree) restore(curr TreeNode, prefix *big.Int, node *DiskNode) TreeNode {
```
nkcr, line 148: I would add a comment about what this function does

core/store/hashtree/binprefix/tree.go
```diff
@@ -280,6 +347,10 @@ func (n *EmptyNode) Delete(key *big.Int, b kv.Bucket) (TreeNode, error) {
 func (n *EmptyNode) Prepare(nonce []byte,
 	prefix *big.Int, b kv.Bucket, fac crypto.HashFactory) ([]byte, error) {
 
+	if len(n.hash) > 0 {
```
nkcr, line 350: I would update the function's comment:
```
// Prepare implements mem.TreeNode. It updates the hash of the node if not 
// already set and returns the digest.
```

core/store/hashtree/binprefix/tree.go
```diff
@@ -452,6 +525,10 @@ func (n *InteriorNode) load(node TreeNode, key *big.Int, bit uint, b kv.Bucket)
 func (n *InteriorNode) Prepare(nonce []byte,
 	prefix *big.Int, b kv.Bucket, fac crypto.HashFactory) ([]byte, error) {
 
+	if len(n.hash) > 0 {
```
nkcr, line 528: same as previous comment, update the function's comment

core/store/hashtree/binprefix/tree.go
```diff
@@ -614,6 +691,10 @@ func (n *LeafNode) Delete(key *big.Int, b kv.Bucket) (TreeNode, error) {
 func (n *LeafNode) Prepare(nonce []byte,
 	prefix *big.Int, b kv.Bucket, fac crypto.HashFactory) ([]byte, error) {
 
+	if len(n.hash) > 0 {
```
nkcr, line 694: same

core/store/hashtree/binprefix/tree.go
```diff
@@ -103,6 +103,73 @@ func (t *Tree) Len() int {
 	return counter
 }
 
+// Load scans the bucket for leafs to insert them in the tree. It will then
+// persist the tree to restore the memory depth limit.
+func (t *Tree) Load(bucket kv.Bucket) error {
```
nkcr, line 108: less elegant, but easier to understand
```suggestion
func (t *Tree) FillFrom(bucket kv.Bucket) error {
```

core/store/hashtree/binprefix/mod.go
Gilthoniel: `Update` => `CalculateRoot`
`Load` => `LaodFromDisk`

That might help a bit.

core/store/hashtree/binprefix/mod.go
Gilthoniel: See the rename above.

core/store/hashtree/binprefix/tree.go
Gilthoniel: No that's actually a good name.

nkcr (approved):

PR #115, opened 2020-09-25
Core: use a thresholded collective signing

Gilthoniel:
This switches from the flat collective signing to a threshold one for the integration tests.

### DIFF
--- a/cosi/threshold/mod.go
+++ b/cosi/threshold/mod.go
@@ -39,8 +39,10 @@ func ByzantineThreshold(n int) int {
 // CoSi is an implementation of the cosi.CollectiveSigning interface that is
 // using streams to parallelize the work.
 type CoSi struct {
-	mino      mino.Mino
-	signer    crypto.AggregateSigner
+	mino   mino.Mino
+	signer crypto.AggregateSigner
+	// Stores the cosi.Threshold function. It will always contain a valid
+	// function by construction.
 	threshold atomic.Value
 }
 
--- a/cosi/threshold/mod.go
+++ b/cosi/threshold/mod.go
@@ -1,35 +1,60 @@
 package threshold
 
 import (
+	"sync/atomic"
+
+	"go.dedis.ch/dela"
 	"go.dedis.ch/dela/cosi"
+	"go.dedis.ch/dela/cosi/threshold/types"
 	"go.dedis.ch/dela/crypto"
 	"go.dedis.ch/dela/mino"
 	"golang.org/x/xerrors"
 )
 
-// Threshold is a function that returns the threshold to reach for a given n.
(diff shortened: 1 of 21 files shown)


cosi/threshold/mod.go
```diff
@@ -47,15 +72,23 @@ func (c *CoSi) GetPublicKeyFactory() crypto.PublicKeyFactory {
...
 // factory.
 func (c *CoSi) GetVerifierFactory() crypto.VerifierFactory {
-	return verifierFactory{factory: c.signer.GetVerifierFactory()}
+	return types.NewThresholdVerifierFactory(c.signer.GetVerifierFactory())
+}
+
+// SetThreshold implements cosi.CollectiveSigning. It sets a new threshold
+// function.
+func (c *CoSi) SetThreshold(fn cosi.Threshold) {
+	if fn == nil {
```
nkcr, line 87: why would fn be nil ?

cosi/threshold/mod.go
```diff
@@ -1,35 +1,60 @@
...
 // CoSi is an implementation of the cosi.CollectiveSigning interface that is
 // using streams to parallelize the work.
 type CoSi struct {
-	mino   mino.Mino
-	signer crypto.AggregateSigner
-
-	Threshold Threshold
+	mino      mino.Mino
+	signer    crypto.AggregateSigner
+	threshold atomic.Value
```
nkcr, line 44: I would  comment that this MUST contain only a type `cosi.Threshold` since you don't do the check in the actor.

cosi/threshold/types/mod.go
```diff
@@ -149,6 +151,14 @@ func (s *Signature) Equal(o crypto.Signature) bool {
 	return ok && other.agg.Equal(s.agg) && bytes.Equal(s.mask, other.mask)
 }
 
+// String implements fmt.Stringer. It returns a string representation of the
+// signature.
+func (s *Signature) String() string {
+	mask := new(big.Int).SetBytes(s.mask)
+
+	return fmt.Sprintf("thres[%b]:%s", mask, s.agg)
```
nkcr, line 159: why just not `fmt.Sprintf("thres[%b]:%s", s.mask, s.agg)` ?

cosi/threshold/mod.go
Gilthoniel: No specific reason but that's a defensive programming approach as it will panic if the type is different.

cosi/threshold/types/mod.go
Gilthoniel: Unfortunately, it prints in an array style like `[0000] [0010]` and I don't like that ;)

nkcr (approved):

PR #116, opened 2020-09-25
Fake package cleaning

Gilthoniel:
This cleans the fake error to use a random value every time the unit tests are run so that it prevents hardcoded values. It also cleans the "oops" errors to use the fake one.


Gilthoniel:
This will probably conflict with previous PRs so it should not be merged right away but after the others.

internal/testing/fake/crypto.go
```diff
@@ -0,0 +1,388 @@
+package fake
+
+import (
+	"hash"
+
+	"go.dedis.ch/dela/crypto"
+	"go.dedis.ch/dela/serde"
+)
+
+// PublicKeyFactory is a fake implementation of a public key factory.
```
nkcr, line 10: In an ideal world we would keep the `- implements ...` convention, same for the functions. What do you think ?

Gilthoniel: Yes, you're right, I wanted to do it but forgot to update the comments.

nkcr (approved):
I'm gonna miss the `xerrors.New("oops")` 😢

PR #120, opened 2020-09-28
Cosipbft: backoff on failure but keep running

Gilthoniel:
This changes the behaviour of the main loop so that it keeps trying the round instead of completely failing.

### DIFF
--- a/core/ordering/cosipbft/mod.go
+++ b/core/ordering/cosipbft/mod.go
@@ -3,6 +3,7 @@ package cosipbft
 import (
 	"context"
 	"fmt"
+	"math"
 	"time"
 
 	"go.dedis.ch/dela"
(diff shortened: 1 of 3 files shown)


internal/testing/fake/mod.go
```diff
@@ -1111,3 +1114,41 @@ func (f AccessControlFactory) Deserialize(serde.Context, []byte) (serde.Message,
...
+		case <-time.After(t):
+			writer.Close()
+		}
+	}()
+
+	go func() {
+		defer cb()
+		defer close(done)
+
+		data := make([]byte, 1024)
```
nkcr, line 1139: just hope that your message is never cut between chunks of 1024 bytes ;)

core/ordering/cosipbft/mod.go
```diff
@@ -374,7 +379,13 @@ func (s *Service) main() error {
 			cancel()
 
 			if err != nil {
-				return xerrors.Errorf("round failed: %v", err)
+				if calculateBackoff(backoff+1) < RoundMaxWait {
```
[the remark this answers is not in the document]
Gilthoniel: Yes because you don't want to reach an infinite timeout but if for instance two many nodes are offline for a long time, it will eventually catch up when they're back online.

internal/testing/fake/mod.go
Gilthoniel: well the log would be too long anyway in that case, but that's for testing so the value can change.

nkcr (approved):

PR #123, opened 2020-09-29
Pool: improve memory management

Gilthoniel:
This improves the memory management of the transaction pool as the history was growing without any limit. It now uses a filter to get rid of invalid transactions.

### DIFF
--- a/core/txn/mod.go
+++ b/core/txn/mod.go
@@ -15,6 +15,7 @@ type Transaction interface {
 	// GetID returns the unique identifier for the transaction.
 	GetID() []byte
 
+	// GetNonce returns the sequence number of the transaction.
 	GetNonce() uint64
 
 	// GetIdentity returns the identity that created the transaction.
--- a/core/txn/mod.go
+++ b/core/txn/mod.go
@@ -15,7 +15,8 @@ type Transaction interface {
 	// GetID returns the unique identifier for the transaction.
 	GetID() []byte
 
-	// GetNonce returns the sequence number of the transaction.
+	// GetNonce returns the nonce of the transaction which corresponds to the
+	// sequence number of a unique identity.
 	GetNonce() uint64
 
 	// GetIdentity returns the identity that created the transaction.
--- a/core/txn/pool/gatherer.go
+++ b/core/txn/pool/gatherer.go
@@ -1,25 +1,60 @@
 package pool
 
 import (
+	"bytes"
 	"context"
-	"fmt"
+	"sort"
 	"sync"
 
+	"go.dedis.ch/dela/core/access"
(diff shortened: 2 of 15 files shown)


core/validation/mod.go
```diff
@@ -30,17 +31,30 @@ type Data interface {
 	GetTransactionResults() []TransactionResult
 }
 
+// DataFactory is the factory for validation data.
 type DataFactory interface {
 	serde.Factory
 
 	DataOf(serde.Context, []byte) (Data, error)
 }
```
nkcr, line 40: comment ?

core/txn/mod.go
```diff
@@ -15,6 +15,7 @@ type Transaction interface {
 	// GetID returns the unique identifier for the transaction.
 	GetID() []byte
 
+	// GetNonce returns the sequence number of the transaction.
```
nkcr, line 18: somehow confusing. We already have a nonce that is associated with an identity in the validation package. Maybe here it should be called `GetSeqNum()` ?

core/txn/pool/gatherer.go
```diff
@@ -1,38 +1,84 @@
...
-const KeyMaxLength = 32
+// DefaultIdentitySize is the default size defined for each identity to store
+// transactions.
+const DefaultIdentitySize = 10
 
-// Key is the key of a transaction. By definition, it expects that the
-// identifier of the transaction is no more than 32 bytes long.
-type Key [KeyMaxLength]byte
+// Transactions is a sortable list of transactions.
+type transactions []txn.Transaction
```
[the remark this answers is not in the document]
Gilthoniel: Oh right there is one public, I always miss it in the doc!

core/txn/mod.go
Gilthoniel: Hm actually the comment is unclear, because it's the sequence number of the transaction *for a given identity*

Gilthoniel: The validation package is using this one so they don't differ from each other.

nkcr: Oh right, makes sense then. I just got confused by the comment

core/txn/pool/gatherer.go
```diff
@@ -41,18 +93,28 @@ type item struct {
...
-	set     map[Key]txn.Transaction
-	history map[Key]struct{}
+
+	limit      int
+	queue      []item
+	validators []Filter
+
+	// A string key is generated for each unique identity, which will have its
+	// own list of transactions.
+	txs map[string]transactions
```
nkcr, line 110: For extra clarity I would explain why we need to store the Txs by identities, instead of having a simple slice or bag of Txs. This is something I'm wondering now

nkcr (approved):

PR #124, opened 2020-09-29
ViewChange: check only one tx per block

Gilthoniel:
This improves the execution service to include previous transactions
so that view change can make sure only one is accepted per block.

### DIFF
--- a/core/execution/mod.go
+++ b/core/execution/mod.go
@@ -5,6 +5,12 @@ import (
 	"go.dedis.ch/dela/core/txn"
 )
 
+// Step is a context of execution.
+type Step struct {
+	Previous []txn.Transaction
+	Current  txn.Transaction
+}
+
 // Result is the result of a transaction execution.
 type Result struct {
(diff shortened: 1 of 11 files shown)


core/execution/mod.go
```diff
@@ -5,6 +5,12 @@ import (
 	"go.dedis.ch/dela/core/txn"
 )
 
+// Step is a context of execution.
```
nkcr, line 8:
```suggestion
// Step is a context of execution. It allows for example a smart contract to execute a given transaction knowing what previous transactions have already been accepted and executed in a block. 
```

nkcr (commented):
I'm wondering, how is that currently implemented in Byzcoin?

Gilthoniel:
In Byzcoin, the whole view change is done quite differently so it's inside that it checks the roster has only one change (and it also rotate for instance). But for instance for the config, I think you can send multiple transactions in one block.

nkcr (approved):

PR #125, opened 2020-09-30
Minogrpc: certificates storage on disk

Gilthoniel:
This implements a persistent storage for the certificates.

Also: store the certificate private key to the disk.

### DIFF
--- a/core/ordering/cosipbft/controller/mod.go
+++ b/core/ordering/cosipbft/controller/mod.go
@@ -1,10 +1,11 @@
 package controller
 
 import (
-	"encoding/base64"
-	"os"
+	"encoding"
 	"path/filepath"
 	"time"
+
+	"go.dedis.ch/dela/crypto"
 
 	"go.dedis.ch/dela/cli"
 	"go.dedis.ch/dela/cli/node"
@@ -22,8 +23,8 @@ import (
 	"go.dedis.ch/dela/core/txn/signed"
 	"go.dedis.ch/dela/core/validation/simple"
 	"go.dedis.ch/dela/cosi/threshold"
-	"go.dedis.ch/dela/crypto"
 	"go.dedis.ch/dela/crypto/bls"
+	"go.dedis.ch/dela/crypto/loader"
 	"go.dedis.ch/dela/mino"
 	"go.dedis.ch/dela/mino/gossip"
 	"go.dedis.ch/dela/serde/json"
@@ -32,11 +33,19 @@ import (
 
 const privateKeyFile = "private.key"
 
-type minimal struct{}
+func blsSigner() encoding.BinaryMarshaler {
+	return bls.NewSigner()
+}
+
+type minimal struct {
+	signerFn func() encoding.BinaryMarshaler
+}
 
 // NewMinimal creates a new minimal controller for cosipbft.
 func NewMinimal() node.Initializer {
-	return minimal{}
+	return minimal{
+		signerFn: blsSigner,
+	}
 }
 
 func (minimal) SetCommands(builder node.Builder) {
@@ -82,31 +91,31 @@ func (minimal) SetCommands(builder node.Builder) {
 	sub.SetAction(builder.MakeAction(rosterAddAction{}))
 }
 
-func (minimal) OnStart(flags cli.Flags, inj node.Injector) error {
-	var m mino.Mino
-	err := inj.Resolve(&m)
+func (m minimal) OnStart(flags cli.Flags, inj node.Injector) error {
+	var onet mino.Mino
+	err := inj.Resolve(&onet)
 	if err != nil {
 		return xerrors.Errorf("injector: %v", err)
 	}
 
-	signer, err := loadOrCreateSigner(flags.Path("config"))
+	signer, err := m.getSigner(flags)
 	if err != nil {
 		return xerrors.Errorf("signer: %v", err)
 	}
 
-	cosi := threshold.NewCoSi(m, signer)
+	cosi := threshold.NewCoSi(onet, signer)
 	cosi.SetThreshold(threshold.ByzantineThreshold)
 
 	exec := baremetal.NewExecution()
 	access := darc.NewService(json.NewContext())
 
-	rosterFac := authority.NewFactory(m.GetAddressFactory(), cosi.GetPublicKeyFactory())
+	rosterFac := authority.NewFactory(onet.GetAddressFactory(), cosi.GetPublicKeyFactory())
 	cosipbft.RegisterRosterContract(exec, rosterFac, access)
 
 	txFac := signed.NewTransactionFactory()
 	vs := simple.NewService(exec, txFac)
 
-	pool, err := poolimpl.NewPool(gossip.NewFlat(m, txFac))
+	pool, err := poolimpl.NewPool(gossip.NewFlat(onet, txFac))
 	if err != nil {
 		return xerrors.Errorf("pool: %v", err)
 	}
@@ -114,13 +123,13 @@ func (minimal) OnStart(flags cli.Flags, inj node.Injector) error {
 	var db kv.DB
 	err = inj.Resolve(&db)
 	if err != nil {
-		return err
+		return xerrors.Errorf("injector: %v", err)
 	}
 
 	tree := binprefix.NewMerkleTree(db, binprefix.Nonce{})
 
 	param := cosipbft.ServiceParam{
-		Mino:       m,
+		Mino:       onet,
 		Cosi:       cosi,
 		Validation: vs,
 		Access:     access,
@@ -142,7 +151,7 @@ func (minimal) OnStart(flags cli.Flags, inj node.Injector) error {
 	}
 
 	blockFac := types.NewBlockFactory(vs.GetFactory())
-	csFac := authority.NewChangeSetFactory(m.GetAddressFactory(), cosi.GetPublicKeyFactory())
+	csFac := authority.NewChangeSetFactory(onet.GetAddressFactory(), cosi.GetPublicKeyFactory())
 	linkFac := types.NewLinkFactory(blockFac, cosi.GetSignatureFactory(), csFac)
 
 	blocks := blockstore.NewDiskStore(db, linkFac)
@@ -192,55 +201,33 @@ func (minimal) OnStop(inj node.Injector) error {
 	return nil
 }
 
-func loadOrCreateSigner(cfg string) (crypto.AggregateSigner, error) {
-	path := filepath.Join(cfg, privateKeyFile)
+func (m minimal) getSigner(flags cli.Flags) (crypto.AggregateSigner, error) {
+	loader := loader.NewFileLoader(filepath.Join(flags.Path("config"), privateKeyFile))
 
-	_, err := os.Stat(path)
-	if os.IsNotExist(err) {
-		signer := bls.NewSigner()
-
-		data, err := signer.MarshalBinary()
-		if err != nil {
-			return nil, xerrors.Errorf("failed to marshal signer: %v", err)
-		}
-
-		file, err := os.Create(path)
-		if err != nil {
-			return nil, xerrors.Errorf("failed to create file: %v", err)
-		}
-
-		defer file.Close()
-
-		enc := base64.NewEncoder(base64.StdEncoding, file)
-		defer enc.Close()
-
-		_, err = enc.Write(data)
-		if err != nil {
-			return nil, xerrors.Errorf("while writing to file: %v", err)
-		}
-
-		return signer, nil
+	signerdata, err := loader.LoadOrCreate(generator{newFn: m.signerFn})
+	if err != nil {
+		return nil, xerrors.Errorf("while loading: %v", err)
 	}
 
-	file, err := os.Open(path)
+	signer, err := bls.NewSignerFromBytes(signerdata)
 	if err != nil {
-		return nil, xerrors.Errorf("failed to open file: %v", err)
-	}
-
-	defer file.Close()
-
-	dec := base64.NewDecoder(base64.StdEncoding, file)
-
-	buffer := make([]byte, 128)
-	n, err := dec.Read(buffer)
-	if err != nil {
-		return nil, xerrors.Errorf("while reading file: %v", err)
-	}
-
-	signer, err := bls.NewSignerFromBytes(buffer[:n])
-	if err != nil {
-		return nil, xerrors.Errorf("failed to unmarshal signer: %v", err)
+		return nil, xerrors.Errorf("while unmarshaling: %v", err)
 	}
 
 	return signer, nil
+}
+
+type generator struct {
+	newFn func() encoding.BinaryMarshaler
+}
+
+func (g generator) Generate() ([]byte, error) {
+	signer := g.newFn()
+
+	data, err := signer.MarshalBinary()
+	if err != nil {
+		return nil, xerrors.Errorf("failed to marshal signer: %v", err)
+	}
+
+	return data, nil
 }
--- a/core/ordering/cosipbft/controller/mod.go
+++ b/core/ordering/cosipbft/controller/mod.go
@@ -111,9 +111,10 @@ func (minimal) OnStart(flags cli.Flags, inj node.Injector) error {
 		return xerrors.Errorf("pool: %v", err)
 	}
 
-	db, err := kv.New(filepath.Join(flags.String("config"), "test.db"))
+	var db kv.DB
+	err = inj.Resolve(&db)
 	if err != nil {
-		return xerrors.Errorf("db: %v", err)
+		return err
 	}
 
 	tree := binprefix.NewMerkleTree(db, binprefix.Nonce{})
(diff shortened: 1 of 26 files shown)


mino/minogrpc/certs/disk_test.go
```diff
@@ -0,0 +1,197 @@
...
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"go.dedis.ch/dela/core/store/kv"
+	"go.dedis.ch/dela/internal/testing/fake"
+	"go.dedis.ch/dela/mino"
+)
+
+func TestDiskStore_Store(t *testing.T) {
```
pierluca, line 15: I would recommend splitting the failing cases into separate tests.

mino/minogrpc/certs/disk_test.go
```diff
@@ -0,0 +1,197 @@
...
+	require.NotNil(t, cert)
+
+	other := NewDiskStore(db, fake.AddressFactory{})
+
+	cert, err = other.Load(fake.NewAddress(0))
+	require.NoError(t, err)
+	require.NotNil(t, cert)
+	cert, err = other.Load(fake.NewAddress(12))
+	require.NoError(t, err)
+	require.NotNil(t, cert)
```
pierluca, line 41: This test does not verify (if I understand correctly) that we get the "right" certificate.
I would create two certs and verify that the store doesn't (for some reason) mix up certificates 0 and 12.
Basically, ignoring the "address" argument would still make this part of the test pass, as I understand it.

mino/minogrpc/certs/disk_test.go
```diff
@@ -0,0 +1,197 @@
...
+	require.EqualError(t, err, fake.Err("while updating db: bucket"))
+
+	memdb := fake.NewInMemoryDB()
+	memdb.SetBucket(store.bucket, fake.NewBadWriteBucket())
+	store.db = memdb
+	err = store.Store(fake.NewAddress(0), cert)
+	require.EqualError(t, err, fake.Err("while updating db: while writing"))
+}
+
+func TestDiskStore_Load(t *testing.T) {
```
pierluca, line 57: same same, I would split in smaller, clearly named tests

mino/minogrpc/certs/disk_test.go
```diff
@@ -0,0 +1,197 @@
...
+	memdb := fake.NewInMemoryDB()
+	memdb.SetBucket(store.bucket, bucket)
+	store.db = memdb
+	_, err = store.Load(fake.NewAddress(0))
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "certificate malformed: asn1: ")
+}
+
+func TestDiskStore_Delete(t *testing.T) {
+	db, clean := makeDb(t)
```
pierluca, line 102: same same, I would split in smaller, clearly named tests

mino/minogrpc/certs/disk_test.go
```diff
@@ -0,0 +1,197 @@
+package certs
```
pierluca, line 1: these tests depend a lot on the textual information contained in the errors, effectively treating those as constants.

As they're somewhat subject to change (e.g. if we want the error to report the address we were trying to access), could we use constants or sentinel errors to achieve the same objective? I may be missing some pros and cons and am open to discussing it.

core/ordering/cosipbft/controller/mod_test.go
```diff
@@ -25,31 +25,65 @@ func TestMinimal_OnStart(t *testing.T) {
 	dir, err := ioutil.TempDir(os.TempDir(), "dela-test-")
```
pierluca, line 25: This is a single test function covering many cases.
Consider splitting in multiple separate test functions

core/ordering/cosipbft/controller/mod_test.go
```diff
@@ -25,31 +25,65 @@ func TestMinimal_OnStart(t *testing.T) {
...
 
-	m := NewMinimal()
+	m := NewMinimal().(minimal)
 
 	fset := make(node.FlagSet)
 	fset["config"] = dir
 
 	inj := node.NewInjector()
 	inj.Inject(fake.Mino{})
+	inj.Inject(db)
```
pierluca, line 40: Cool. I like injectors :-P

pierluca (changes requested):
I haven't finished the review, I still need to look at the files listed below. So far so good, it's very elegant. Just a few minor things here and there :)

mino/minogrpc/controller/actions_test.go 
mino/minogrpc/controller/mod.go 
mino/minogrpc/controller/mod_test.go 
mino/minogrpc/mod.go 
mino/minogrpc/mod_test.go 
mino/minogrpc/server.go 
mino/minogrpc/server_test.go

internal/testing/fake/store.go
```diff
@@ -49,3 +52,145 @@ func (snap *InMemorySnapshot) Delete(key []byte) error {
...
+func (db *InMemoryDB) Update(fn func(tx kv.WritableTx) error) error {
+	return fn(dbTx{buckets: db.buckets, err: db.err})
+}
+
+// Close implements kv.DB.
+func (db *InMemoryDB) Close() error {
+	return db.err
+}
+
+type dbTx struct {
```
nkcr, line 113: `// - implements kv.WritableTx`

mino/minogrpc/certs/disk.go
```diff
@@ -0,0 +1,198 @@
...
+		return xerrors.Errorf("while updating db: %v", err)
+	}
+
+	s.InMemoryStore.Store(addr, cert)
+
+	return nil
+}
+
+// Load implements certs.Storage. It first tries to read the certificate from
+// the cache, then from the disk.
```
nkcr, line 72:
```suggestion
// the cache, then from the disk. Returns nil if not found in both.
```

core/ordering/cosipbft/controller/mod.go
```diff
@@ -82,44 +91,45 @@ func (minimal) SetCommands(builder node.Builder) {
 	sub.SetAction(builder.MakeAction(rosterAddAction{}))
 }
 
-func (minimal) OnStart(flags cli.Flags, inj node.Injector) error {
-	var m mino.Mino
-	err := inj.Resolve(&m)
+func (m minimal) OnStart(flags cli.Flags, inj node.Injector) error {
+	var onet mino.Mino
```
nkcr, line 95: Arg 😱, I never thought I would see that word in Dela 😄

internal/testing/fake/store.go
```diff
@@ -49,3 +52,145 @@ func (snap *InMemorySnapshot) Delete(key []byte) error {
...
+	return &InMemoryDB{
+		buckets: make(map[string]*Bucket),
+	}
+}
+
+// NewBadDB returns a new database that will return an error inside the
+// transactions.
+func NewBadDB() *InMemoryDB {
+	db := NewInMemoryDB()
+	db.err = fakeErr
```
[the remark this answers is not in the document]
Gilthoniel: Yes, you're right.

mino/minogrpc/certs/disk.go
```diff
@@ -0,0 +1,198 @@
...
+	"errors"
+
+	"go.dedis.ch/dela/core/store/kv"
+	"go.dedis.ch/dela/mino"
+	"golang.org/x/xerrors"
+)
+
+var certBucket = []byte("certificates")
+
+var errInterrupt = errors.New("interrupted")
```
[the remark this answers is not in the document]
Gilthoniel: The idea was to not declare a variable every time the function is called, as I'm not sure how the compiler optimizes that kind of situations.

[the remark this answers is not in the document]
pierluca: I'm generally against premature optimization, but you know more than I do about this.
Are we on the code hot path?

mino/minogrpc/controller/mod.go
```diff
@@ -1,26 +1,43 @@
...
 	"golang.org/x/xerrors"
 )
 
+const certKeyName = "cert.key"
+
 // Minimal is an initializer with the minimum set of commands.
 //
 // - implements node.Initializer
-type minimal struct{}
+type minimal struct {
```
pierluca, line 30: I have a bit of an issue with name.
`minimal` is an adjective, it doesn't quite describe what this is/does.
Can you think of a better name?

mino/minogrpc/controller/mod.go
```diff
@@ -1,26 +1,43 @@
...
+	"go.dedis.ch/dela/core/store/kv"
+	"go.dedis.ch/dela/crypto/loader"
 	"go.dedis.ch/dela/mino"
 	"go.dedis.ch/dela/mino/minogrpc"
+	"go.dedis.ch/dela/mino/minogrpc/certs"
 	"go.dedis.ch/dela/mino/router/tree"
 	"golang.org/x/xerrors"
 )
 
+const certKeyName = "cert.key"
```
pierluca, line 25: is there any need for this to be configurable?
I'm not sure when this implementation would be used.

mino/minogrpc/controller/mod_test.go
```diff
@@ -20,11 +26,20 @@ func TestMinimal_Build(t *testing.T) {
 }
 
 func TestMinimal_OnStart(t *testing.T) {
-	minimal := NewMinimal()
+	dir, err := ioutil.TempDir(os.TempDir(), "minogrpc")
```
pierluca, line 29: I know you're going to hate me at this point, but I'd still favor small tests with descriptive names :)

mino/minogrpc/controller/mod_test.go
```diff
@@ -34,14 +49,46 @@ func TestMinimal_OnStart(t *testing.T) {
...
 }
 
 func TestMinimal_OnStop(t *testing.T) {
+	dir, err := ioutil.TempDir(os.TempDir(), "minogrpc")
+	require.NoError(t, err)
+
+	db, err := kv.New(filepath.Join(dir, "test.db"))
+	require.NoError(t, err)
+
+	defer os.RemoveAll(dir)
```
pierluca, line 84: I would move this `defer` before  the `kv.New` test, in case that crashes.

mino/minogrpc/controller/mod.go
```diff
@@ -87,7 +104,25 @@ func (m minimal) OnStart(ctx cli.Flags, inj node.Injector) error {
...
+
+	key, err := m.getKey(ctx)
+	if err != nil {
+		return xerrors.Errorf("cert private key: %v", err)
+	}
+
+	opts := []minogrpc.Option{
+		minogrpc.WithStorage(certs),
+		minogrpc.WithCertificateKey(key, key.Public()),
+	}
```
pierluca, line 123: Just a design question: is this a common pattern in the code base ?

If not, what do you think about using a Fluent API? For configuration it's very natural. It might look like:

```
	opts := minogrpc.Options().
                   WithStorage(certs).
                   WithCertificateKey(key, key.Public())
```
and then you pass only one parameter, `opts`.

You could then call it with `opts.Apply(tmpl)` and all the details are hidden away from the library users. 

Then again, maybe this is actually just very idiomatic Go and I'm coming with baggage :-)

mino/minogrpc/mod_test.go
```diff
@@ -16,7 +16,9 @@ import (
 func TestMinogrpc_New(t *testing.T) {
 	addr := ParseAddress("127.0.0.1", 3333)
```
pierluca, line 17: yes, I'm going to favor smaller tests :-P

mino/minogrpc/server.go
```diff
@@ -386,52 +386,76 @@ type overlay struct {
...
-	addrFactory mino.AddressFactory, ctx serde.Context) (*overlay, error) {
-
-	cert, err := makeCertificate()
-	if err != nil {
-		return nil, xerrors.Errorf("failed to make certificate: %v", err)
-	}
-
-	meBytes, err := me.MarshalText()
+func newOverlay(tmpl minoTemplate) (*overlay, error) {
+	meBytes, err := tmpl.me.MarshalText()
```
pierluca, line 398: as a newcomer to the codebase, `me` is *extremely* confusing. 
Can we name this field with a more explicit name? e.g. `myAddr` or similar.

mino/minogrpc/server.go
```diff
@@ -386,52 +386,76 @@ type overlay struct {
 	router      router.Router
 	connMgr     session.ConnectionManager
 	addrFactory mino.AddressFactory
+	secret      interface{}
+	public      interface{}
 
 	// Keep a text marshalled value for the overlay address so that it's not
 	// calculated for each request.
 	meStr string
 }
```
pierluca, line 395: It wouldn't hurt to document the fields of the overlay struct, again, as a newcomer :)

mino/minogrpc/server_test.go
```diff
@@ -525,9 +558,19 @@ func TestConnManager_Acquire(t *testing.T) {
...
 	require.EqualError(t, err, "certificate for 'fake.Address[1]' not found")
 
+	mgr.certs = fakeCerts{errLoad: fake.GetError(), counter: fake.NewCounter(1)}
+	_, err = mgr.Acquire(fake.NewAddress(0))
+	require.EqualError(t, err, fake.Err("while loading own cert"))
+
+	mgr.certs = certs
 	mgr.conns = make(map[mino.Address]*grpc.ClientConn)
 	mgr.certs.Delete(mgr.me)
 	_, err = mgr.Acquire(dst.GetAddress())
```
pierluca, line 576: Sorry for repeating myself, but I would split the failing cases in separate tests

pierluca (commented):
Finished the second part of the review.

A few discussion points, but overalls I'm very happy with how much I can follow, which at this point means the code is clean, (mostly) well named and documented :)

mino/minogrpc/certs/disk_test.go
Gilthoniel: To discuss, the codebase follows this pattern for errors. It's out of scope of this PR to change, but I guess we can talk about pros and cons.

core/store/kv/controller/mod.go
```diff
@@ -0,0 +1,44 @@
...
+type minimal struct{}
+
+func NewMinimal() node.Initializer {
+	return minimal{}
+}
+
+func (m minimal) SetCommands(builder node.Builder) {}
+
+func (m minimal) OnStart(flags cli.Flags, inj node.Injector) error {
+	db, err := kv.New(filepath.Join(flags.String("config"), "test.db"))
```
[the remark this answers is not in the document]
Gilthoniel: Right, I prefer using a constant name for now, if you find it useful in the future, that's left to you.

core/store/kv/controller/mod.go
```diff
@@ -0,0 +1,44 @@
...
+	if err != nil {
+		return xerrors.Errorf("db: %v", err)
+	}
+
+	inj.Inject(db)
+
+	return nil
+}
+
+func (m minimal) OnStop(inj node.Injector) error {
```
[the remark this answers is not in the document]
Gilthoniel: You have a base config folder which contains the different files created. The tests are then responsible to clean that folder. I don't think it makes sense to offer a parameter to destroy the files when the daemon stops, I just don't see why you would do that.

mino/minogrpc/controller/mod.go
Gilthoniel: More than to be configurable, it's simply good practice to use constants when the code is using a value like this. That's what I learnt, but there's probably different mindsets.

pierluca: Oh, I agree with you entirely.
I was just asking if this was a hardcoded constant or if later on it might need to be configured.

mino/minogrpc/controller/mod.go
Gilthoniel: I'm following this design pattern: https://www.sohamkamani.com/golang/options-pattern/

Two main reasons:
1. gRPC does too
2. I hate the formatting of chained functions in Go

Like always that's open to discussion but like you said it's a general pattern in the code base, so out of scope of this PR.

mino/minogrpc/controller/mod.go
Gilthoniel: Then the answer is yes, that's only a hardcoded constant, which will probably evolve to a default name and an option to set a different name.

mino/minogrpc/controller/mod.go
pierluca: Good enough for me :-) 

And yes, function chaining in Go looks horrible...

mino/minogrpc/certs/disk.go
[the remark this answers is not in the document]
Gilthoniel: It's more a cold path at this moment but I would treat that like the string constants. Unfortunately you can't have a constant error in Go, but it's the same as declaring a constant when you use an actual number or string.

nkcr (approved):

mino/minogrpc/controller/mod_test.go
pierluca: @Gilthoniel thumb'd up but not changed. is it on purpose?

pierluca (approved):
LGTM, thanks !

PR #126, opened 2020-10-01
Cosipbft: fix a synchronization issue with the tree

Gilthoniel:
This fixes a synchronization issue between the tree and the latest block which is failing the test from time to time.

Also: improve blocksync

### DIFF
--- a/core/ordering/cosipbft/blockstore/mem.go
+++ b/core/ordering/cosipbft/blockstore/mem.go
@@ -170,5 +170,16 @@ type observer struct {
 }
 
 func (obs observer) NotifyCallback(evt interface{}) {
-	obs.ch <- evt.(types.BlockLink)
+	for {
+		select {
+		case obs.ch <- evt.(types.BlockLink):
+			// Event went through.
+			return
+		default:
+			// A event is waiting to be read from the listener but the watcher
+			// needs to push this event forward. The channel is thus drained,
+			// and populated again with the most recent event.
+			<-obs.ch
+		}
+	}
 }
--- a/core/ordering/cosipbft/blockstore/mem.go
+++ b/core/ordering/cosipbft/blockstore/mem.go
@@ -170,6 +170,8 @@ type observer struct {
 }
 
 func (obs observer) NotifyCallback(evt interface{}) {
+	// TODO: use a non-blocking queue to prevent a event from blocking when the
+	// channel is busy.
 	for {
 		select {
 		case obs.ch <- evt.(types.BlockLink):
--- a/core/ordering/cosipbft/blockstore/mod.go
+++ b/core/ordering/cosipbft/blockstore/mod.go
@@ -21,7 +21,7 @@ type TreeCache interface {
 	// of the cache alongside a function to unlock the cache. It allows one to
 	// delay a set while fetching associated data. The function returned must be
 	// called.
-	GetWithLock() (hashtree.Tree, func())
+	GetWithLock() (tree hashtree.Tree, unlock func())
 
 	Set(hashtree.Tree)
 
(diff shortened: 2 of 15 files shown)


core/ordering/cosipbft/blockstore/mem.go
```diff
@@ -170,5 +170,16 @@ type observer struct {
 }
 
 func (obs observer) NotifyCallback(evt interface{}) {
-	obs.ch <- evt.(types.BlockLink)
+	for {
```
nkcr, line 173: that feels strange that an observer can actually misses some events because he wasn't quick enough. If that doesn't pose a problem it should be well mentioned.

core/ordering/cosipbft/blockstore/mod.go
```diff
@@ -18,11 +17,19 @@ var ErrNoBlock = errors.New("no block")
 type TreeCache interface {
 	Get() hashtree.Tree
 
+	// GetWithLock implements blockstore.TreeCache. It returns the current value
+	// of the cache alongside a function to unlock the cache. It allows one to
+	// delay a set while fetching associated data. The function returned must be
+	// called.
+	GetWithLock() (hashtree.Tree, func())
```
nkcr, line 24:
```suggestion
	GetWithLock() (t hashtree.Tree, unlock func())
```

core/ordering/cosipbft/blockstore/mod.go
```diff
@@ -18,11 +17,19 @@ var ErrNoBlock = errors.New("no block")
...
 	Set(hashtree.Tree)
 
-	// SetAndLock sets the tree and acquires the cache lock until the wait group
-	// is done which allows an external caller to delay the release of the lock.
-	SetAndLock(hashtree.Tree, *sync.WaitGroup)
+	// SetWithLock implements blockstore.TreeCache. It sets the tree while
+	// holding the lock and returns a function to unlock it. It allows one to
+	// prevent an access until associated data is updated. The function returned
+	// must be called.
+	SetWithLock(hashtree.Tree) func()
```
nkcr, line 32:
```suggestion
	SetWithLock(hashtree.Tree) (unlock func())
```

core/ordering/cosipbft/blocksync/default.go
```diff
@@ -110,43 +117,63 @@ func (s defaultSync) Sync(ctx context.Context, players mino.Players, cfg Config)
...
 	// the context.
-	soft := map[mino.Address]struct{}{}
-	hard := map[mino.Address]struct{}{}
+	wg := sync.WaitGroup{}
+	wg.Add(1)
+	once := sync.Once{}
+
+	// The synchronization is run in background so that it continues even after
+	// the threshold is reached, which allow other nodes to complete a catch up
+	// whthe round is performing.
```
nkcr, line 126:
```suggestion
	// while the round is performing.
```

internal/testing/fake/mino.go
```diff
@@ -285,7 +285,7 @@ func (r *Receiver) Recv(context.Context) (mino.Address, serde.Message, error) {
 
 	// In the case there are no more messages to read we return the last one
```
nkcr, line 286:
```suggestion
	// In the case there are no more messages to read we return nil
```

core/ordering/cosipbft/blockstore/mem.go
Gilthoniel: Your feeling is right, that was a dirty solution to fix a blocking issue (like you planned it with the remark on the watcher basically), but I will make a PR to use a non-blocking buffer like mino. Adding a TODO for the moment.

core/ordering/cosipbft/blocksync/default.go
Gilthoniel: my laptop is slower than my fingers.. !

nkcr (approved):

PR #128, opened 2020-10-01
Cosipbft: recover from a block propagation failure

Gilthoniel:
This improves the round procedure to handle cases where a block is
committed but the leader failed to propagate it to followers.


Gilthoniel:
The address implementation of Minogrpc had to move to the session package because of cycling imports.

core/ordering/cosipbft/pbft/mod.go
```diff
@@ -186,33 +188,55 @@ func (m *pbftsm) GetViews() map[mino.Address]View {
...
+	if uint16(index) != m.round.leader {
+		return id, xerrors.Errorf("'%v' is not the leader", from)
+	}
+
+	// Check the state after verifying that the proposal comes from the right
+	// leader.
+	if m.state == PrepareState || m.state == CommitState {
+		// The leader should only propose one block, therefore the accepted
+		// proposal identifier is sent back, whatever the input is.
+		return id, nil
```
nkcr, line 232: then what happens to the block? The leader will wait for this block to be committed, which will reach a timeout?

core/ordering/cosipbft/pbft/mod.go
```diff
@@ -226,12 +250,7 @@ func (m *pbftsm) Commit(id types.Digest, sig crypto.Signature) error {
...
 	defer m.Unlock()
 
-	if m.state == CommitState {
-		// The state machine is already committed to a proposal.
-		return nil
-	}
-
-	if m.state != PrepareState {
+	if m.state != PrepareState && m.state != CommitState {
 		return xerrors.Errorf("mismatch state %v != %v", m.state, PrepareState)
```
nkcr, line 254: need an error message update

mino/minoch/mod.go
```diff
@@ -12,15 +12,20 @@ import (
 	"go.dedis.ch/dela/serde/json"
 )
 
+// Filter is a function called for any request to an RPC which will drop it if
+// it returns false.
+type Filter func(mino.Request) bool
```
nkcr, line 17: isn't it a bit expensive to add all this mechanism just for the sake of testing ?

mino/minogrpc/session/addr.go
```diff
@@ -0,0 +1,145 @@
...
+
+// NewAddress creates a new address.
+func NewAddress(host string) Address {
+	return Address{host: host}
+}
+
+// GetDialAddress returns a string formatted to be understood by grpc.Dial()
+// functions.
+func (a Address) GetDialAddress() string {
+	return a.host
```
nkcr, line 46: we actually never check that a.host is valid

mino/minogrpc/session/addr.go
```diff
@@ -0,0 +1,145 @@
...
+
+// FromText implements mino.AddressFactory. It returns an instance of an
+// address from a byte slice.
+func (f AddressFactory) FromText(text []byte) mino.Address {
+	if len(text) == 0 {
+		return Address{}
+	}
+
+	if string(text)[0] == orchestratorCode[0] {
+		return Address{host: string(text)[1:], orchestrator: true}
```
nkcr, line 141: it panics if len(text) == 1

mino/minogrpc/session/addr.go
Gilthoniel: Well if that's the case, it will fail to dial.

mino/minoch/mod.go
Gilthoniel: Minoch has been implemented for the sake of testing, so I don't think so. It's also very cheap as the default behaviour of being empty does add anything to the logic.

core/ordering/cosipbft/pbft/mod.go
Gilthoniel: Yes, if the leader decides to propose a different block, the followers will simply reject it and proceed to a view change after the timeout. Then new leader == new block.

mino/minogrpc/session/addr.go
Gilthoniel: it actually doesn't: https://play.golang.org/p/LYMPrve_bQK

`[2:]` would panic though.

Gilthoniel: I'll add a test anyway.

nkcr (approved):

nkcr:
Can be merged once the conflict is resolved.

Gilthoniel:
I'll wait for #125 to be merged, otherwise it will be pain to rebase it.

PR #130, opened 2020-10-05
Documentation: update flatcosi comments, add example.

Gilthoniel:
This adds technical documentation to the code, as well as examples.

### DIFF
--- a/cosi/flatcosi/mod.go
+++ b/cosi/flatcosi/mod.go
@@ -1,5 +1,5 @@
 // Package flatcosi is a flat implementation of a collective signing so that the
-// orchestrator will contact all of the participants to require their signature
+// orchestrator will contact all the participants to require their signatures
 // and then aggregate them to the final one.
 //
 // Documentation Last Review: 05.10.2020
--- a/cosi/flatcosi/mod.go
+++ b/cosi/flatcosi/mod.go
@@ -1,3 +1,9 @@
+// Package flatcosi is a flat implementation of a collective signing so that the
+// orchestrator will contact all of the participants to require their signature
+// and then aggregate them to the final one.
+//
+// Documentation Last Review: 05.10.2020
+//
 package flatcosi
 
 import (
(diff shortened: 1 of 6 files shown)


cosi/mod.go
```diff
@@ -13,20 +29,21 @@ type Reactor interface {
 	serde.Factory
```
nkcr, line 29: I would explain why we need a Reactor and how it is used

cosi/flatcosi/mod.go
```diff
@@ -1,3 +1,9 @@
+// Package flatcosi is a flat implementation of a collective signing so that the
+// orchestrator will contact all of the participants to require their signature
```
nkcr, line 2:
```suggestion
// orchestrator will contact all the participants to require their signatures
```

cosi/flatcosi/mod.go
```diff
@@ -93,7 +103,8 @@ type flatActor struct {
 	reactor cosi.Reactor
 }
 
-// Sign implements cosi.Actor. It returns the collective signature of the block.
+// Sign implements cosi.Actor. It returns the collective signature of the
+// message if every participant returns its signature.
```
nkcr, line 107: a threshold, no ?

Gilthoniel: No this implementation is oriented to be simple, there is no support for a mask and all that stuff that is needed like in the actual threshold package.

cosi/mod.go
```diff
@@ -24,7 +24,9 @@ import (
 	"go.dedis.ch/dela/serde"
 )
 
-// Reactor is an event handler that demultiplex the events.
+// Reactor is a collective signature event handler. Every participant must react
+// to a incoming signature request from the leader, and this abstraction
```
nkcr, line 28:
```suggestion
// to an incoming signature request from the leader, and this abstraction
```

nkcr (approved):
that would be good if @pierluca can also give his newcomer eye on that.

pierluca (approved):
It took me a little bit of back-and-forth, but frankly, I don't think this can be any clearer.
LGTM :)

PR #131, opened 2020-10-05
Blockstore: use a buffer for watch events

Gilthoniel:
This changes the observer to not drop events when the channel is busy
but fill a buffer that will deliver the events in order progressively.

### DIFF
--- a/core/ordering/cosipbft/blockstore/mem.go
+++ b/core/ordering/cosipbft/blockstore/mem.go
@@ -4,11 +4,17 @@ import (
 	"context"
 	"sync"
 
+	"github.com/rs/zerolog"
+	"go.dedis.ch/dela"
 	"go.dedis.ch/dela/core"
 	"go.dedis.ch/dela/core/ordering/cosipbft/types"
 	"go.dedis.ch/dela/core/store"
 	"golang.org/x/xerrors"
 )
+
+// sizeWarnLimit defines the limit after which a warning is emitted every time
+// the queue keeps growing.
+const sizeWarnLimit = 100
 
 // InMemory is a block store that only stores the block in-memory which means
 // they won't persist.
@@ -166,6 +172,7 @@ func (s *InMemory) WithTx(txn store.Transaction) BlockStore {
 type observer struct {
 	sync.Mutex
 
+	logger  zerolog.Logger
 	buffer  []types.BlockLink
 	running bool
 	closed  bool
@@ -175,7 +182,8 @@ type observer struct {
 func newObserver(ctx context.Context, watcher core.Observable) *observer {
 	ch := make(chan types.BlockLink, 1)
 	obs := &observer{
-		ch: ch,
+		logger: dela.Logger,
+		ch:     ch,
 	}
 
 	watcher.Add(obs)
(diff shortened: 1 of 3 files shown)


core/ordering/cosipbft/blockstore/mem.go
```diff
@@ -165,23 +158,106 @@ func (s *InMemory) WithTx(txn store.Transaction) BlockStore {
...
+
+	select {
+	case obs.ch <- evt.(types.BlockLink):
+
+	default:
+		// The buffer size is not controlled as we assume the event will be read
+		// shortly by the caller.
+		obs.buffer = append(obs.buffer, evt.(types.BlockLink))
+
+		if !obs.running {
```
nkcr, line 215: thanks to the lock, this seems not possible that `obs.running` has changed its value since the previous check in line 202.

core/ordering/cosipbft/blockstore/mem.go
```diff
@@ -165,23 +158,106 @@ func (s *InMemory) WithTx(txn store.Transaction) BlockStore {
...
+// Observer is an observer that can be added to store watcher. It will announce
+// the blocks in order and without blocking the watcher even if the listener is
+// not actively emptying the queue.
+//
+// - implements core.Observer.
 type observer struct {
-	ch chan types.BlockLink
+	sync.Mutex
+
+	working sync.WaitGroup
```
nkcr, line 169: is this wait group necessary ?

Gilthoniel: No that was for the tests, but ended up not necessary.

nkcr (approved):

core/ordering/cosipbft/blockstore/mem.go
```diff
@@ -165,23 +158,100 @@ func (s *InMemory) WithTx(txn store.Transaction) BlockStore {
...
+		obs.buffer = append(obs.buffer, evt.(types.BlockLink))
+		return
+	}
+
+	select {
+	case obs.ch <- evt.(types.BlockLink):
+
+	default:
+		// The buffer size is not controlled as we assume the event will be read
+		// shortly by the caller.
```
pierluca, line 211: this looks reasonable at the moment, but also like the kind of thing that could some day blow up in our face and might take some time to debug.

Even if we're not handling it, can we at least put some conditional WARN logging if the buffer is above a certain (insane) size?

pierluca (changes requested):
Tiny remark, LGTM overalls.

core/ordering/cosipbft/blockstore/mem.go
Gilthoniel: That seems reasonable ;)

core/ordering/cosipbft/blockstore/mem.go
```diff
@@ -4,12 +4,18 @@ import (
...
 
+	"github.com/rs/zerolog"
+	"go.dedis.ch/dela"
 	"go.dedis.ch/dela/core"
 	"go.dedis.ch/dela/core/ordering/cosipbft/types"
 	"go.dedis.ch/dela/core/store"
 	"golang.org/x/xerrors"
 )
 
+// sizwWarnLimit defines the limit after which a warning is emitted every time
```
pierluca, line 15: typo `sizwWarnLimit`

PR #132, opened 2020-10-05
Documentation: thresholded collective signing documentation

Gilthoniel:
This add the technical documentation to the threshold collective signing.

Also: clean some of the tests

### DIFF
--- a/core/ordering/cosipbft/blocksync/default_test.go
+++ b/core/ordering/cosipbft/blocksync/default_test.go
@@ -20,7 +20,6 @@ import (
 	"go.dedis.ch/dela/internal/testing/fake"
 	"go.dedis.ch/dela/mino"
 	"go.dedis.ch/dela/mino/minoch"
-	"go.dedis.ch/dela/serde"
 )
 
 func TestDefaultSync_Basic(t *testing.T) {
@@ -100,7 +99,11 @@ func TestDefaultSync_GetLatest(t *testing.T) {
 }
 
 func TestDefaultSync_Sync(t *testing.T) {
-	rcvr := fake.NewReceiver(types.NewSyncRequest(0), types.NewSyncRequest(0), types.NewSyncAck())
+	rcvr := fake.NewReceiver(
+		fake.NewRecvMsg(fake.NewAddress(0), types.NewSyncRequest(0)),
+		fake.NewRecvMsg(fake.NewAddress(0), types.NewSyncRequest(0)),
+		fake.NewRecvMsg(fake.NewAddress(0), types.NewSyncAck()),
+	)
 	sender := fake.Sender{}
 
 	sync := defaultSync{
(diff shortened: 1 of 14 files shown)


cosi/threshold/types/mod.go
```diff
@@ -1,3 +1,9 @@
+// Package types implements the threshold collective signature and its verifier.
+// It wraps a signature implementation in order to extract the correct
+// aggregated public key.
```
nkcr, line 3: maybe that would be nice to explain why we use a separate package for that.

nkcr (approved):

core/ordering/cosipbft/blocksync/default_test.go
```diff
@@ -182,15 +189,21 @@ func TestHandler_Stream(t *testing.T) {
 	handler.pbftsm = testSM{blocks: handler.blocks}
```
pierluca, line 189: Would it make sense to split this in small tests ? :)

cosi/threshold/handler.go
```diff
@@ -1,56 +1,59 @@
+// This file contains the implementation of the RPC handler.
```
pierluca, line 1: is this comment accurate? It feels - at the very least - a bit incomplete

cosi/threshold/types/mod.go
```diff
@@ -265,10 +279,16 @@ func NewThresholdVerifierFactory(fac crypto.VerifierFactory) crypto.VerifierFact
 	}
 }
 
+// FromAuthority implements crypto.VerifierFactory. It creates a verifier from
+// the authority so that the mask's signature will pick the participants that
+// have participated. The ordering of the authority must be the same.
```
pierluca, line 284: Newbie question:
`The ordering of the authority must be the same.`
what does this mean?

dkg/pedersen/mod_test.go
```diff
@@ -52,15 +52,17 @@ func TestPedersen_Setup(t *testing.T) {
 	_, err = actor.Setup(fakeAuthority, 1)
```
pierluca, line 52: 😇  how do you feel about splitting tests?

pierluca (commented):
Minor remarks. LGTM

core/ordering/cosipbft/blocksync/default_test.go
Gilthoniel: I'm doing this progressively when updating the documentation.

cosi/threshold/handler.go
Gilthoniel: Hm no, it's really only doing this. It implements the Stream handler for the RPC instantiated by the module.

cosi/threshold/types/mod.go
Gilthoniel: The signature mask defines how participated from an ordered list of signers. If the list is shuffled, the aggregate public key will be wrong. The authority abstraction is there to prevent this, but you could still use a different one with a different order.

dkg/pedersen/mod_test.go
Gilthoniel: for DKG, that will be on @nkcr hands, he's the owner of the package !

pierluca (approved):

PR #133, opened 2020-10-05
Documentation: add technical doc to the crypto packages

Gilthoniel:


crypto/mod.go
```diff
@@ -1,3 +1,19 @@
+// Package crypto defines cryptographic primitives shared by the different
+// modules of Dela.
+//
+// It defines the abstraction of a public and a private key and their
+// combination where a private key can create a signature for a given message,
+// while the public key can verify the associated of the those two elements.
```
nkcr, line 6:
```suggestion
// while the public key can verify the association of the those two elements.
```

crypto/mod.go
```diff
@@ -1,3 +1,19 @@
...
+//
+// A signer is a unique entity that possesses both a public and an associated
+// private key. It is assumed that the combination is a unique identity. This
+// abstraction hides the logic a private key.
```
nkcr, line 10:
```suggestion
// abstraction hides the logic of a private key.
```

nkcr (approved):

crypto/bls/mod.go
```diff
@@ -1,3 +1,12 @@
+// Package bls implements the cryptographic primitives using the BLS signature
+// scheme and the BN256 elliptic curve.
```
pierluca, line 2: isn't it an issue that we're still using BLS instead of BDN, considering it's broken?
It's outside the scope of this PR, but if I'm correct, you might want to add a deprecated warning here as well.

Gilthoniel: BLS is not really broken but there is a possible attack with a rogue public key. So depending on the situation, your system can be safe in that matter.

pierluca (approved):

PR #134, opened 2020-10-05
Minogrpc: verify validity of a shared certificate

Gilthoniel:


pierluca (approved):
Seems pretty straightforward :) LGTM

mino/minogrpc/addr.go
```diff
@@ -50,6 +53,16 @@ func (a address) GetDialAddress() string {
 	return a.host
 }
 
+// GetHostname parses the address to extract the hostname.
+func (a address) GetHostname() (string, error) {
+	url, err := url.Parse(fmt.Sprintf("//%s", a.host))
+	if err != nil {
+		return "", xerrors.Errorf("malformed address: %v", err)
```
nkcr, line 60:
```suggestion
		return "", xerrors.Errorf("failed to parse address: %v", err)
```

otherwise you get a: `malformed address: malformed address: ...` error message

mino/minogrpc/server.go
```diff
@@ -122,19 +122,35 @@ func (o overlayServer) Join(ctx context.Context, req *ptypes.JoinRequest) (*ptyp
...
+	from := o.addrFactory.FromText(msg.GetAddress()).(address)
 
 	cert, err := x509.ParseCertificate(msg.GetValue())
 	if err != nil {
 		return nil, xerrors.Errorf("couldn't parse certificate: %v", err)
 	}
 
-	o.certs.Store(from, &tls.Certificate{Leaf: cert})
+	// Make sure the certificate is valid for the public key provided.
+	err = cert.CheckSignature(cert.SignatureAlgorithm, cert.RawTBSCertificate, cert.Signature)
```
nkcr, line 135: apparently its the same as doing `cert.CheckSignatureFrom(cert)`, but it adds few more checks.

Gilthoniel: It does not accept the same certificate as a valid parent.

mino/minogrpc/addr.go
Gilthoniel: If forgot to write the test for the hostname, but after doing it, it seems like the error makes sense.

mino/minogrpc/server.go
Gilthoniel: nevermind, it's a question of configuration.

nkcr (approved):

PR #135, opened 2020-10-06
Documentation: add examples to crypto packages

Gilthoniel:
Also: update comment for the loader

### DIFF
--- a/crypto/common/example_test.go
+++ b/crypto/common/example_test.go
@@ -1,0 +1,86 @@
+package common_test
+
+import (
+	"fmt"
+
+	"go.dedis.ch/dela/crypto/bls"
+	"go.dedis.ch/dela/crypto/common"
+	"go.dedis.ch/dela/crypto/ed25519"
+	"go.dedis.ch/dela/serde/json"
+)
+
+func ExamplePublicKeyFactory_PublicKeyOf_bls() {
+	// BLS is already registered by default
+	factory := common.NewPublicKeyFactory()
+
+	ctx := json.NewContext()
+
+	message := []byte("42")
+
+	signer := bls.NewSigner()
+	publicKey := signer.GetPublicKey()
+
+	signature, err := signer.Sign(message)
+	if err != nil {
+		panic("signing failed: " + err.Error())
+	}
+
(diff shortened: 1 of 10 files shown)


crypto/common/example_test.go
```diff
@@ -0,0 +1,86 @@
+package common_test
+
+import (
+	"fmt"
+
+	"go.dedis.ch/dela/crypto/bls"
```
pierluca, line 6: Kyber's BLS indicates that it's deprecated because it's got security flaws.
We probably shouldn't use it for the example as it may give some bad ideas :) 

May I suggest to replace it with bdn?

pierluca (changes requested):
LGTM on the documentation side, I believe there's a security issue with one of the examples.

crypto/ed25519/example_test.go
```diff
@@ -0,0 +1,23 @@
...
+	message := []byte("42")
+
+	signature, err := signerA.Sign(message)
+	if err != nil {
+		panic("signer failed: " + err.Error())
+	}
+
+	err = signerA.GetPublicKey().Verify(message, signature)
+	if err != nil {
+		panic("invalid signature: " + err.Error())
```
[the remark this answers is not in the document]
Gilthoniel: unfortunately there is no such type. If the signature is invalid, the Kyber error is something like "bls invalid signature" but you have to dig why.

crypto/common/example_test.go
Gilthoniel: Like we discuss, BDN fixes the use of rogue public key which is possible in certain circumstances and BDN is also drawbacks, so it's a matter of deciding. BLS itself works, so the example is not flawed.

pierluca (approved):

crypto/loader/example_test.go
```diff
@@ -0,0 +1,34 @@
...
+	if err != nil {
+		panic("loading key failed: " + err.Error())
+	}
+
+	fmt.Println(string(data))
+
+	// Output: a marshaled private key
+}
+
+type exampleGenerator struct{}
```
nkcr, line 30: -- implements loader.Generator

nkcr (approved):

PR #136, opened 2020-10-06
Memcoin: fix signal issues with Windows during tests

Gilthoniel:
Because the tests were listening for the terminate signal, that was causing some instability. It now only waits for the channel to be closed in the test, and listen for the signal in production.

### DIFF
--- a/cli/node/builder.go
+++ b/cli/node/builder.go
@@ -24,6 +24,7 @@ type cliBuilder struct {
 	daemonFactory DaemonFactory
 	injector      Injector
 	actions       *actionMap
+	enableSignal  bool
 	sigs          chan os.Signal
 	startFlags    []cli.Flag
 	commands      []*cliCommand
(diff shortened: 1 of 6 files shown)


cli/node/builder.go
```diff
@@ -24,6 +24,7 @@ type cliBuilder struct {
 	daemonFactory DaemonFactory
 	injector      Injector
 	actions       *actionMap
+	enableSignal  bool
```
nkcr, line 27: I would explain why we need it, ie. in the case of testing where the notify can cause instability.

cli/node/daemon.go
```diff
@@ -129,6 +129,10 @@ func (d *socketDaemon) handleConn(conn net.Conn) {
 		return
 	}
 
+	dela.Logger.Debug().
```
pierluca, line 132: question: is the debug level off by default?

pierluca (approved):
Tiny question, bother otherwise LGTM

cli/node/daemon.go
Gilthoniel: That's going to be documented but there is an environment variable to enabled them `LLVL=trace go test` so that they can be turned on on-demand. Otherwise it can be confusing when having warnings in the tests (but it's expected).

nkcr (approved):

PR #140, opened 2020-10-06
Documentation: update mino packages documentation

Gilthoniel:
This updates the documentation of the mino packages minus minogrpc and proxy.

Also: minoch new function does not return an error but panic.

### DIFF
--- a/mino/minoch/example_test.go
+++ b/mino/minoch/example_test.go
@@ -11,8 +11,8 @@ import (
 func ExampleRPC_Call() {
 	manager := NewManager()
 
-	minoA := NewMinoch(manager, "A")
-	minoB := NewMinoch(manager, "B")
+	minoA := MustCreate(manager, "A")
+	minoB := MustCreate(manager, "B")
 
 	roster := mino.NewAddresses(minoA.GetAddress(), minoB.GetAddress())
 
--- a/mino/minoch/example_test.go
+++ b/mino/minoch/example_test.go
@@ -1,0 +1,96 @@
+package minoch
+
+import (
+	"context"
+	"fmt"
+
+	"go.dedis.ch/dela/mino"
+	"go.dedis.ch/dela/serde"
+)
+
+func ExampleRPC_Call() {
+	manager := NewManager()
+
+	minoA := NewMinoch(manager, "A")
+	minoB := NewMinoch(manager, "B")
+
+	roster := mino.NewAddresses(minoA.GetAddress(), minoB.GetAddress())
+
+	nsA, err := minoA.MakeNamespace("example")
+	if err != nil {
+		panic("namespace failed: " + err.Error())
+	}
+
+	rpcA, err := nsA.MakeRPC("hello", exampleHandler{}, exampleFactory{})
+	if err != nil {
+		panic("rpc failed: " + err.Error())
+	}
+
+	nsB, err := minoB.MakeNamespace("example")
+	if err != nil {
+		panic("namespace failed: " + err.Error())
+	}
+
+	_, err = nsB.MakeRPC("hello", exampleHandler{}, exampleFactory{})
+	if err != nil {
+		panic("rpc failed: " + err.Error())
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	msg := exampleMessage{value: "Hello World!"}
+
+	resps, err := rpcA.Call(ctx, msg, roster)
+	if err != nil {
+		panic("call failed: " + err.Error())
+	}
+
+	for resp := range resps {
+		reply, err := resp.GetMessageOrError()
+		if err != nil {
+			panic("error in response: " + err.Error())
+		}
+
+		fmt.Println(reply.(exampleMessage).value)
+	}
+
+	// Output: Hello World!
+	// Hello World!
+}
+
+// exampleHandler is an RPC handler example.
+//
+// - implements mino.Handler
+type exampleHandler struct {
+	mino.UnsupportedHandler
+}
+
+// Process implements mino.Handler. It returns the message received.
+func (exampleHandler) Process(req mino.Request) (serde.Message, error) {
+	return req.Message, nil
+}
+
+// exampleMessage is an example of a message.
+//
+// - implements serde.Message
+type exampleMessage struct {
+	value string
+}
+
+// Serialize implements serde.Message. It returns the value contained in the
+// message.
+func (m exampleMessage) Serialize(serde.Context) ([]byte, error) {
+	return []byte(m.value), nil
+}
+
+// exampleFactory is an example of a factory.
+//
+// - implements serde.Factory
+type exampleFactory struct{}
+
+// Deserialize implements serde.Factory. It returns the message using data as
+// the inner value.
+func (exampleFactory) Deserialize(ctx serde.Context, data []byte) (serde.Message, error) {
+	return exampleMessage{value: string(data)}, nil
+}
--- a/mino/minoch/mod.go
+++ b/mino/minoch/mod.go
@@ -1,5 +1,16 @@
-// Package minoch is an implementation of MINO that is using channels and a
+// Package minoch is an implementation of Mino that is using channels and a
 // local manager to exchange messages.
+//
+// Because it is using only Go channels to communicate, this implementation can
+// only be used by multiple instances in the same process. Its usage is purely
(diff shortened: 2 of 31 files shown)


mino/minoch/mod.go
```diff
@@ -29,7 +43,7 @@ type Minoch struct {
 }
 
 // NewMinoch creates a new instance of a local Mino instance.
-func NewMinoch(manager *Manager, identifier string) (*Minoch, error) {
+func NewMinoch(manager *Manager, identifier string) *Minoch {
```
nkcr, line 46: to avoid later trouble, maybe that would be better to have a `minoch.MustCreate()` function that panics, and let this function return an error for the cases we want to be safe ? (I hear you saying "minoch is for testing purposes anyway" 😄)

Gilthoniel: No that's actually a good compromise. It's just that it's annoying to have to handle those kind of errors (like the namespace and the rpc) that can happen only during instantiation if you're really doing something wrong. In that case I think it's fair to take a panic in the face.

Gilthoniel: What do you think about a mino.MustCreateRPC(name string, ...) mino.RPC function and similarly for the namespace ?

nkcr: mmm `MakeNamespace` can easily return an error with a malformed namespace, so I'm not convinced. `MakeRPC`, on the other hand, has no reason to return an error, at least with minogrpc, so `MustCreateRPC` seems like a good idea.

Gilthoniel: For the record, we've decided to do as @nkcr said for Minoch and change the namespace and rpc functions to panic if necessary.

mino/minoch/rpc.go
```diff
@@ -30,14 +37,14 @@ type RPC struct {
 	filters []Filter
 }
 
-// Call sends the message to all participants and gather their reply. The
-// context is ignored in the scope of channel communication as there is no
-// blocking I/O. The response channel will receive n responses for n players and
-// be closed eventually.
+// Call implements mino.RPC. It sends the message to all participants and gather
```
nkcr, line 40:
```suggestion
// Call implements mino.RPC. It sends the message to all participants and gathers
```

mino/minoch/rpc.go
```diff
@@ -30,14 +37,14 @@ type RPC struct {
...
+// their reply. The context is ignored in the scope of channel communication as
```
nkcr, line 41:
```suggestion
// their replies. The context is ignored in the scope of channel communication as
```

mino/mod.go
```diff
@@ -48,19 +86,18 @@ type Players interface {
...
+	// The response channel must be closed after every request ended in a
+	// result, either a reply or an error.
+	Call(ctx context.Context, req serde.Message, players Players) (<-chan Response, error)
 
-// Receiver is an interface to provide primitives to receive messages from
-// recipients.
-type Receiver interface {
-	Recv(context.Context) (Address, serde.Message, error)
+	// Stream is a persistent request that will be closed only when the
+	// orchestrator is done or an error occured.
```
nkcr, line 99:
```suggestion
	// orchestrator is done or an error occured, or the context is done.
```

mino/minoch/example_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+func ExampleRPC_Call() {
+	manager := NewManager()
+
+	minoA := NewMinoch(manager, "A")
+	minoB := NewMinoch(manager, "B")
+
+	roster := mino.NewAddresses(minoA.GetAddress(), minoB.GetAddress())
+
+	nsA, err := minoA.MakeNamespace("example")
```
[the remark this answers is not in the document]
Gilthoniel: that will be soon segmentA :p

mino/minoch/example_test.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	// Output: Hello World!
+	// Hello World!
+}
+
+// exampleHandler is an RPC handler example.
+//
+// - implements mino.Handler
+type exampleHandler struct {
```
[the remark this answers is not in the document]
Gilthoniel: There will be one example with Minogrpc.

nkcr (approved):

mino/router/tree/example_test.go
```diff
@@ -0,0 +1,65 @@
...
+	addrB := session.NewAddress("127.0.0.1:3000")
+
+	players := mino.NewAddresses(addrA, addrB)
+
+	table, err := routerA.New(players)
+	if err != nil {
+		panic("routing table failed: " + err.Error())
+	}
+
+	handshake := table.Prelude(addrB)
```
pierluca, line 47: Probably out of scope for this review:

I feel this example highlights that "Prelude" is not necessarily the best name for the interface.

It took me a while (and reading the rest of the documentation) to understand the intention of this example, as "Prelude" really doesn't say *what* this is doing, but only *when* it should be done.

mino/router/tree/example_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+	// Send the handshake to the address B..
+
+	routerB := NewRouter(session.AddressFactory{})
+
+	tableB, err := routerB.TableOf(handshake)
```
pierluca, line 53: Similarly out of scope:

`TableOf` has a similar issue.
What we're doing here (if I understand correctly) is more like a `GenerateTableFrom` no?

mino/mod.go
```diff
@@ -48,19 +86,18 @@ type Players interface {
...
-	// Send sends the message to all the addresses. It returns a channel that
-	// will be populated with errors coming from the network layer if the
-	// message cannot be sent. The channel must be closed after the message has
-	// been/failed to be sent.
-	Send(msg serde.Message, addrs ...Address) <-chan error
-}
+// RPC is an abstraction of a distributed Remote Procedure Call.
+type RPC interface {
+	// Call is a basic request to one or multiple distant peers. It directly
+	// contacts all the players and thus expect a reasonable number of peers.
```
pierluca, line 92:
```suggestion
	// contacts all the players and thus expects a reasonable number of peers.
```

mino/mod.go
```diff
@@ -48,19 +86,18 @@ type Players interface {
...
+	//
+	// The response channel must be closed after every request ended in a
+	// result, either a reply or an error.
```
pierluca, line 95: Am I understanding correctly? The phrasing is a bit ambiguous

```suggestion
	// The response channel must be closed once the request ends with a
	// result, either a reply or an error.
```

mino/mod.go
```diff
@@ -48,19 +86,18 @@ type Players interface {
...
+	Call(ctx context.Context, req serde.Message, players Players) (<-chan Response, error)
 
-// Receiver is an interface to provide primitives to receive messages from
-// recipients.
-type Receiver interface {
-	Recv(context.Context) (Address, serde.Message, error)
+	// Stream is a persistent request that will be closed only when the
+	// orchestrator is done or an error occured, or the context is done.
```
pierluca, line 99:
```suggestion
	// orchestrator is done or an error has occured, or the context is done.
```

mino/mod.go
```diff
@@ -86,19 +123,23 @@ type Response interface {
...
-	// The response channel must be closed after every request ended in a
-	// result, either a reply or an error.
-	Call(ctx context.Context, req serde.Message, players Players) (<-chan Response, error)
+// Sender is an abstraction to send messages to a stream in the context of a
+// distributed RPC.
+type Sender interface {
+	// Send sends the message to all the addresses. It returns a channel that
+	// will be populated with errors coming from the network layer if the
+	// message cannot be sent. The channel must be closed after the message has
+	// been/failed to be sent.
```
pierluca, line 132:
```suggestion
	// been sent or failed to be.
```

pierluca (commented):
Minor things here and there. A few questions too.

Thanks for the documentation !

mino/router/tree/example_test.go
Gilthoniel: Oh I would definitely use a different name, I just couldn't find one !

mino/router/tree/example_test.go
Gilthoniel: Yes, I can change to this.

nkcr (approved):

pierluca (approved):
LGTM

PR #141, opened 2020-10-07
Documentation: update minogrpc documentation

Gilthoniel:
Also: fix events watch in controller to be before the tx is sent

### DIFF
--- a/core/ordering/cosipbft/controller/action.go
+++ b/core/ordering/cosipbft/controller/action.go
@@ -137,27 +137,9 @@ func (rosterAddAction) Execute(ctx node.Context) error {
 		return xerrors.Errorf("injector: %v", err)
 	}
 
-	roster, err := srvc.GetRoster()
+	tx, err := prepareRosterTx(ctx, srvc)
 	if err != nil {
-		return xerrors.Errorf("failed to read roster: %v", err)
-	}
-
-	addr, pubkey, err := decodeMember(ctx, ctx.Flags.String("member"))
-	if err != nil {
-		return xerrors.Errorf("failed to decode member: %v", err)
-	}
-
-	cset := authority.NewChangeSet()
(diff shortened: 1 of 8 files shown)


mino/minogrpc/mod.go
```diff
@@ -224,12 +237,9 @@ func (m *Minogrpc) postCheckClose() error {
 	return nil
 }
 
-// MakeNamespace implements Mino. It creates a new Minogrpc
-// struct that has the specified namespace. This namespace is further used to
-// scope newly created RPCs. There can be multiple namespaces. If there is
-// already a namespace, then the new one will be concatenated leading to
-// namespace1/namespace2. A namespace can not be empty an should match
-// [a-zA-Z0-9]+
+// MakeNamespace implements mino.Mino. It returns a instance that extends the
```
nkcr, line 240:
```suggestion
// MakeNamespace implements mino.Mino. It returns an instance that extends the
```

nkcr (approved):

core/ordering/cosipbft/controller/action.go
```diff
@@ -166,31 +166,35 @@ func (rosterAddAction) Execute(ctx node.Context) error {
...
-		err := waitTx(srvc, wait, tx)
-		if err != nil {
-			return xerrors.Errorf("wait: %v", err)
-		}
-	}
-
-	return nil
-}
-
-func waitTx(srvc Service, wait time.Duration, tx txn.Transaction) error {
```
pierluca, line 185: I'm not sure that joining these two functions again makes for a smoother read.
I'd split them the other way around though: one doing all the initialisation, which Execute() would call, and then doing the wait part inline. 
Either ways, this is okay. We can always refactor later if need be.

pierluca (approved):
LGTM

PR #142, opened 2020-10-07
Documentation: update minogrpc packages

Gilthoniel:
Also: add minogrpc examples

### DIFF
--- a/mino/minogrpc/certs/example_test.go
+++ b/mino/minogrpc/certs/example_test.go
@@ -1,0 +1,40 @@
+package certs_test
+
+import (
+	"fmt"
+
+	"go.dedis.ch/dela/mino/minogrpc"
+	"go.dedis.ch/dela/mino/minogrpc/certs"
+	"go.dedis.ch/dela/mino/minogrpc/session"
+	"go.dedis.ch/dela/mino/router/tree"
+)
+
+func ExampleStorage_Fetch() {
+	router := tree.NewRouter(minogrpc.NewAddressFactory())
+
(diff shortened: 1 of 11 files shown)


mino/minogrpc/session/queue.go
```diff
@@ -38,13 +46,15 @@ func newNonBlockingQueue() *NonBlockingQueue {
 	}
 }
 
-// Channel implements minogrpc.Queue. It returns the message channel.
+// Channel implements session.Queue. It returns the message channel and it
+// should be read as much as possible to avoid messages to be dropped.
```
nkcr, line 50: that's very mysterious  🧙‍♂️

nkcr (approved):

mino/minogrpc/certs/example_test.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	store := certs.NewInMemoryStore()
+
+	digest, err := store.Hash(m.GetCertificate())
+	if err != nil {
+		panic("certificate digest failed: " + err.Error())
+	}
+
+	err = store.Fetch(m.GetAddress().(session.Address), digest)
```
pierluca, line 27: Naive question, I haven't looked into this too much: 
is the cast expected and needed, or is there a better alternative through the API ?

pierluca (approved):

mino/minogrpc/certs/example_test.go
Gilthoniel: When used internally by minogrpc, it knowns the address as session.Address, but for the example I have to get the address from the instance which returns the interface. I could create the address from the host value though.

PR #143, opened 2020-10-07
Documentation: update serde doc and add examples

Gilthoniel:
Also: fix unstable test with channel synchronization


core/ordering/cosipbft/blockstore/mem.go
```diff
@@ -258,20 +263,20 @@ func (obs *observer) pushAndWait() {
...
-		case <-obs.ch:
-		default:
-		}
+	// Drain message in transit to close the channel properly.
+	select {
+	case <-obs.ch:
+	default:
 	}
 
 	close(obs.ch)
```
nkcr, line 277: I may be wrong, but if there are multiple messages in the buffer, the `pushAndWait` will still try to send those messages, but the chan will be closed.
..

Ok, this works because `len(obs.buffer) == 0` will return true, I get it.

serde/registry/example_test.go
```diff
@@ -0,0 +1,80 @@
+package registry_test
```
nkcr, line 1: I had the hope it would make things clearer but honestly I still struggle. example_test.go and registry_test.go seems to do the same, but with slightly different means, and this is not clear what the differences are. I'm still lost.

Gilthoniel: As discussed, I merged the examples and will do a following PR to add another format to demonstrate the use case.

nkcr (approved):

pierluca (approved):
LGTM

PR #155, opened 2020-10-12
Fix: daemon socket communication

Gilthoniel:
This improves the communication between a client and the daemon so that error are detected as such and make the command fails. This should improve the connection with the daemon under Windows.

### DIFF
--- a/cli/node/daemon.go
+++ b/cli/node/daemon.go
@@ -10,12 +10,20 @@ import (
 	"sync"
 	"time"
 
+	"github.com/rs/zerolog"
 	"go.dedis.ch/dela"
 	"go.dedis.ch/dela/cli"
 	"golang.org/x/xerrors"
 )
 
 const ioTimeout = 30 * time.Second
+
+// event is the structure sent over the connection between the client and the
+// daemon and vice-versa using a JSON encoding.
+type event struct {
+	Err   bool
+	Value string
+}
 
 // SocketClient opens a connection to a unix socket daemon to send commands.
 //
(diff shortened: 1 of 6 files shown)


cli/node/daemon_test.go
```diff
@@ -18,7 +19,7 @@ import (
 func TestSocketClient_Send(t *testing.T) {
```
pierluca, line 19: split into small tests? :)

cli/node/daemon_test.go
```diff
@@ -121,6 +116,83 @@ func TestSocketDaemon_Listen(t *testing.T) {
...
+	defer daemon.Close()
+
+	conn, err := net.DialTimeout("unix", daemon.socketpath, 1*time.Second)
+	require.NoError(t, err)
+	require.NoError(t, conn.Close())
+}
+
+func TestSocketDaemon_ConnClosedFromClient_Listen(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		// This test does not work on Windows.
```
pierluca, line 144: Consider adding more details to include if it's "by design" or not, and why.

Otherwise my natural tendency would be to remove this and analyze why as a sort of bug hunt.

cli/node/daemon.go
```diff
@@ -10,13 +10,21 @@ import (
 	"sync"
```
pierluca, line 10: so clean. 
/me like

pierluca (approved):
LGTM, minor remarks

cli/node/daemon_test.go
Gilthoniel: I refactored the test to use a fake connection instead. The issue is Windows having a different behaviour than Linux in terms of when/how it returns the errors. And this test is only there to test one log.

cli/node/daemon_test.go
Gilthoniel: I need to write an example for this package, so I'll do that at the same time.

PR #156, opened 2020-10-12
Mino: namespace refactoring

Gilthoniel:
This changes the way namespaces are defined. It now considers the namespace as something that already exists and you simply access it and it renames namespace to segment.

### DIFF
--- a/mino/minoch/example_test.go
+++ b/mino/minoch/example_test.go
@@ -11,30 +11,13 @@ import (
 func ExampleRPC_Call() {
 	manager := NewManager()
 
-	minoA := MustCreate(manager, "A")
-	minoB := MustCreate(manager, "B")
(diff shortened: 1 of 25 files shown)


mino/minoch/example_test.go
```diff
@@ -11,30 +11,13 @@ import (
 func ExampleRPC_Call() {
 	manager := NewManager()
 
-	minoA := MustCreate(manager, "A")
```
nkcr, line 14: yeah, I see now why you wanted to remove the return error, this example got a nice diet 😄

nkcr (approved):
I love starting the day with a PR that removes more code than it adds 😀

pierluca (approved):
LGTM. Love it actually :)

PR #159, opened 2020-10-13
Documentation: add ledger section to the design doc

Gilthoniel:


Gilthoniel:
The cosipbft part is not following a intro/development/conclusion, I was just throwing the different thoughts I had about the design.

docs/cosipbft.md
```diff
@@ -0,0 +1,76 @@
...
+
+1. Any of the state can transition to _ViewChange_ (other than (V) itself).
+2. After a view change, it either transition to _Initial_ if no candidate has
+   been comitted, otherwise to _Commit_.
+3. The usual transitions when nothing goes wrong is _Initial_ to _Prepare_ to
+   _Commit_ and back to _Initial_ after the block is finalized.
+
+The point (2.) is explained because when a candidate is committed on at least
+one of the participant, it means that a threshold has also committed because the
+signature exists! The system will have to finalize this candidate, even if a new
```
nkcr, line 69: one can't create a cosi on its own ?

docs/ledger.md
```diff
@@ -0,0 +1,75 @@
...
+
+Each participant of a distributed ledger needs to collect a bunch of
+transactions when it is trying to create a block. The pool is there to provide
+that service, so that clients can send their transactions to one of the pool of
+the distributed system while the ordering services wait for enough transactions
+to be discovered to start a new block.
+
+The purpose is to offer a single entry point for the client and then the system
+will take care of spreading the transactions so that any reachable member will
+at some point learn about it.
```
nkcr, line 41: And if you don't trust the system, better contact multiple pools on different nodes it one wants to be sure the tx won't be censored. Right ?

docs/ledger.md
```diff
@@ -0,0 +1,75 @@
...
+which could allow a malicious client to execute a smart contract in a incorrect
+way. The validation service makes sure the result of a transaction execution
+only updated what is allowed.
+
+Finally, because it is the validation that defines what the transaction looks
+like, it also provides a manager that will help to create and sign transactions.
+For instance, in the case of Ethereum, a client needs to use the correct nonce
+for the transaction to be accepted.
+
+## Ordering Service
```
nkcr, line 61: something we haven't discuss it how one can retrieve the state of a smart contract? With Byzcoin you can do a `proof.KeyValue` and de-serialize the smart contract's content. How would that work with Dela ?

nkcr (approved):

docs/cosipbft.md
Gilthoniel: Not sure I get the question, but unlike bitcoin, we don't want mini-fork so when a block is committed, it must be the next one. Anyone can start a cosi, but in the context of pbft, only the requests from the leader get a reply.

docs/ledger.md
Gilthoniel: Yes that's a good point

docs/ledger.md
Gilthoniel: That really depends on the execution environment but you have multiple solutions.

docs/cosipbft.md
```diff
@@ -0,0 +1,76 @@
...
+| N | ---------> | I | -----------> | P | ------------> | C | <-----
+-----            ----- <-----       -----               -----      |
+  |                |         |        |                   |        |
+  |                |         |        |                   |        |
+  |                |       -----      |                   |        |
+  -----------------------> | V | <-------------------------        |
+                           -----                                   |
+                             |                                     |
+                             ---------------------------------------
+```
```
pierluca, line 59: For completeness, I would number the transitions and provide a detailed description as to what event would trigger the change.

docs/cosipbft.md
```diff
@@ -0,0 +1,76 @@
...
+
+1. Any of the state can transition to _ViewChange_ (other than (V) itself).
+2. After a view change, it either transition to _Initial_ if no candidate has
+   been comitted, otherwise to _Commit_.
+3. The usual transitions when nothing goes wrong is _Initial_ to _Prepare_ to
+   _Commit_ and back to _Initial_ after the block is finalized.
```
pierluca, line 65: what it means to "go wrong" may be different for each state. It'd be good to document that on a high level.

docs/cosipbft.md
pierluca: Based on @nkcr question and my own confusion, I feel it might be good to clarify this paragraph.

Tell me if I understand correctly:

The point (2.) is explained because we may be in a view-change state while a threshold of participants accept a new block. Furthermore, reaching a threshold is a prerequisite to a candidate being committed on any participant. As such, if a candidate has been committed on at least one of the participants, we know that the signature exists. The system will have to finalize this candidate, even if a new leader tries a different one. A different candidate will be refused anyway by the participants committed to the other.

docs/ledger.md
```diff
@@ -0,0 +1,75 @@
...
+left to the ledger implementation to define how values are stored and
+identified. The transaction is created with a targeted execution environment
+alongside the input parameters.
+
+The first abstraction which allows one to propagate transactions is defined as a
+pool that can work independetly from the other abstractions. In fact, you could
+run a distributed system that only shares transactions.
+
+The second is a service that accepts a transaction and can return if it will be
+included in the next block, or if it will be rejected. Another aspect defines
```
pierluca, line 21: Another aspect of what? 
Is this a "Furthermore, it defines..."?

docs/ledger.md
pierluca: @Gilthoniel @nkcr something to discuss this afternoon? :)

pierluca (commented):
Thanks for the good work. A few questions here and there.

docs/ledger.md
nkcr: yes, good idea 🍺

docs/cosipbft.md
Gilthoniel: it was mostly if the round expires, but it's confusing.

docs/cosipbft.md
Gilthoniel: I also fixed the None state that goes to Prepare actually.

pierluca (approved):
LGTM

PR #160, opened 2020-10-13
Documentation: update cli comments

Gilthoniel:


cli/node/example_test.go
```diff
@@ -0,0 +1,101 @@
+package node
```
nkcr, line 1: 👌

nkcr (approved):

cli/node/example_test.go
```diff
@@ -0,0 +1,101 @@
...
+// daemon side.
+type Hello interface {
+	SayTo(name string)
+}
+
+type simpleHello struct{}
+
+func (simpleHello) SayTo(name string) {
+	fmt.Printf("Hello, %s!", name)
+}
```
pierluca, line 48: this logic is not demonstrated in the example though and is just dead (example) code, or am I missing something?
I'm a tiny bit confused

pierluca (commented):
Tiny remark on the example, but overalls 👍 👍 👍

cli/node/example_test.go
Gilthoniel: To demonstrate, I need to start a daemon and that's quite complex in the example as it should be an independent command but I wanted to show how to write the controller.

pierluca (approved):
LGTM

cli/node/example_test.go
pierluca: Fair enough :)

PR #161, opened 2020-10-14
Serde: add XML to the example

Gilthoniel:
This demonstrates how different formats can be used by simply switching the context.

Needs #143

### DIFF
--- a/serde/example_test.go
+++ b/serde/example_test.go
@@ -7,36 +7,48 @@ import (
 	"go.dedis.ch/dela/serde"
 	"go.dedis.ch/dela/serde/json"
 	"go.dedis.ch/dela/serde/registry"
+	"go.dedis.ch/dela/serde/xml"
 )
 
-func ExampleMessage_Serialize_json() {
+func ExampleMessage_Serialize() {
 	// Register a JSON format engine for the message type.
 	exampleRegistry.Register(serde.FormatJSON, exampleJSONFormat{})
+	// Register a Gob format engine for the message type.
+	exampleRegistry.Register(serde.FormatXML, exampleXMLFormat{})
 
 	msg := exampleMessage{
 		value: 42,
 	}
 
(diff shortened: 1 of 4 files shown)


serde/example_test.go
```diff
@@ -130,3 +144,45 @@ func (exampleJSONFormat) Decode(ctx serde.Context, data []byte) (serde.Message,
...
+	if err != nil {
+		return nil, err
+	}
+
+	msg := exampleMessage{
+		value: m.Value,
+	}
+
+	return msg, nil
+}
```
pierluca, line 188: why do this and not use the same `exampleMessageEncoded` (merging `exampleMessageXML` and `exampleMessageJSON`) with an attribute  ``` `json:"value" xml:"value"` ``` ?

pierluca (commented):
LGTM, though we may be able to avoid some repetition ?

serde/xml/mod.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"go.dedis.ch/dela/serde"
+)
+
+// xmlEngine is a context engine that uses the XML encoding. See encoding/xml.
+//
+// - implements serde.ContextEngine
+type xmlEngine struct{}
+
+// NewContext returns a new serde context that is using the Gob encoding.
```
nkcr, line 18:
```suggestion
// NewContext returns a new serde context that is using the XML encoding.
```

serde/example_test.go
Gilthoniel: While it would work, the goal is to show you can split entirely the logic of the formats because in actual use cases, there is a format implementation in an isolated package.

nkcr (approved):

serde/example_test.go
pierluca: Fair enough. Thanks for the clarification :)

PR #164, opened 2020-12-07
Fixes wrong marshalling of int flags

### LINKED ISSUE #163: Int flags received as float on the deamon
We use `json.Marshal(fset)` to send the flags from the cli to the deamon. This operation transforms `Int` values to the `float` type, which then makes the retrieving of `Int` flags impossible.

This is due to JSON marshaling: https://play.golang.org/p/BcnbnaB3cxy

nkcr:
Checks for float flags and convert them back to int if possible. Updates the test.

fixes #163


pierluca (approved):
LGTM

PR #165, opened 2020-12-11
Refactors cli/node to have an independent urfave implementation

nkcr:
This change cleans up a bit the cli/node package. 
I extracted the urfave implementation to ease the creation of other CLIs, like the one I'm preparing to use the crypto package.

### DIFF
--- a/cli/urfave/mod.go
+++ b/cli/urfave/mod.go
@@ -13,7 +13,7 @@ import (
 //
 // - implements cli.Builder
 type Builder struct {
-	commands map[string]*cmdBuilder
+	commands []*cmdBuilder
 	name     string
 	action   cli.Action
 	flags    []cli.Flag
@@ -24,10 +24,9 @@ type Builder struct {
 // provides the global flags available from all the commands/subcommands.
 func NewBuilder(name string, action cli.Action, flags ...cli.Flag) cli.Builder {
 	return &Builder{
-		name:     name,
-		commands: make(map[string]*cmdBuilder),
-		action:   action,
-		flags:    flags,
+		name:   name,
+		action: action,
+		flags:  flags,
 	}
 }
 
@@ -48,9 +47,9 @@ func (b Builder) Build() cli.Application {
 // SetCommand implements cli.Builder.
 func (b *Builder) SetCommand(name string) cli.CommandBuilder {
 	cmd := &cmdBuilder{
-		subcommands: make(map[string]*cmdBuilder),
+		name: name,
 	}
-	b.commands[name] = cmd
+	b.commands = append(b.commands, cmd)
 
 	return cmd
 }
@@ -59,10 +58,11 @@ func (b *Builder) SetCommand(name string) cli.CommandBuilder {
 //
 // - implements cli.CommandBuilder
 type cmdBuilder struct {
+	name        string
 	description string
 	action      cli.Action
 	flags       []ucli.Flag
-	subcommands map[string]*cmdBuilder
+	subcommands []*cmdBuilder
 }
 
 // SetDescription implements cli.CommandBuilder.
@@ -83,9 +83,9 @@ func (b *cmdBuilder) SetAction(action cli.Action) {
 // SetSubCommand implements cli.CommandBuilder.
 func (b *cmdBuilder) SetSubCommand(name string) cli.CommandBuilder {
 	builder := &cmdBuilder{
-		subcommands: make(map[string]*cmdBuilder),
+		name: name,
 	}
-	b.subcommands[name] = builder
+	b.subcommands = append(b.subcommands, builder)
 
 	return builder
 }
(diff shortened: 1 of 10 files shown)


cli/urfave/mod.go
```diff
@@ -0,0 +1,165 @@
+// Package urfave provides a cli builder implementation based on the urfave/cli
+// library.
+package urfave
```
pierluca, line 3: why name this after urfave, considering that all the exported types, functions, etc. are NOT urfave-dependent? As a matter of fact, if I understand correctly, from this package's consumer's perspective, this package's dependence on urfave is an implementation detail.
We could swap it out for another dependency and the user wouldn't notice.

could we name the package after its function? would `builder` be suited ?

cli/urfave/mod.go
```diff
@@ -0,0 +1,165 @@
...
+		Flags:    buildFlags(b.flags),
+	}
+
+	app.Setup()
+
+	return app
+}
+
+// SetCommand implements cli.Builder.
+func (b *Builder) SetCommand(name string) cli.CommandBuilder {
```
pierluca, line 49: I'm a tiny bit conflicted.
In Go, I believe the best practice is to take abstractions as parameters and return concretions, i.e. take in interfaces and return structs. On the other hand, it makes a lot of sense to return that abstraction here, in terms of code eloquence...

cli/urfave/mod.go
```diff
@@ -0,0 +1,165 @@
...
+func (b *cmdBuilder) SetAction(action cli.Action) {
+	b.action = action
+}
+
+// SetSubCommand implements cli.CommandBuilder.
+func (b *cmdBuilder) SetSubCommand(name string) cli.CommandBuilder {
+	builder := &cmdBuilder{
+		subcommands: make(map[string]*cmdBuilder),
+	}
+	b.subcommands[name] = builder
```
pierluca, line 88: shouldn't we do error-checking to verify that `b.subcommands[name]` hasn't already been defined?

cli/urfave/mod_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+	fakeAction := func(flags cli.Flags) error {
+		return nil
+	}
+
+	res = makeAction(fakeAction)
+	require.NotNil(t, res)
+
+	out := res(&ucli.Context{})
+	require.NoError(t, out)
```
pierluca, line 117: the logic can be severely broken without this test failing.
Can we do some "tougher" asserts about `res` and `out` ?

cli/urfave/mod_test.go
```diff
@@ -0,0 +1,118 @@
...
+		cli.IntFlag{
+			Name:     "name4",
+			Usage:    "usage4",
+			Required: true,
+			Value:    1,
+		},
+	}
+
+	out := buildFlags(in)
+	require.Len(t, out, 4)
```
pierluca, line 93: you might want to add a few assertions here with respect to what's really outputted.

cli/urfave/mod_test.go
```diff
@@ -0,0 +1,118 @@
...
+func TestSetCommand(t *testing.T) {
+	builder := NewBuilder("test", nil)
+
+	builder.SetCommand("first")
+	builder.SetCommand("second")
+
+	app := builder.Build().(*ucli.App)
+
+	// urfave adds by default the -help command. This is why we expect 3.
+	require.Len(t, app.Commands, 3)
```
pierluca, line 34: can we partially replace the comment with some asserts about there being a "first", "second" and "help" ? :)

pierluca (changes requested):
Few things here and there, nothing major :)

cli/urfave/mod.go
nkcr: Even if the exposed functions are not urfave dependent, we name it after its underlying implementation. A user is then free to choose its implementation based on the implementation details, which can matter, like with could provide a `fake` cli. 
This is how it's done in Dela. This is why we, for example, have `minogrpc` and `minoch`, which both provides the same exported types but with different implementation details. Could say the same with the ordering module: we have a `pow (proof-of-work)` one and a `cosipbft` one.

cli/urfave/mod.go
nkcr: That would be useful only if we needed to use that function from within the package and we needed to handle a `cmdBuilder`, which is not the case. An external user would see the abstraction anyway.

cli/urfave/mod.go
pierluca: OK, I can accept this.
Could we at least have 'cli' in the package name, like for mino? It'd make the using code more readable (ex. `urcli.NewBuilder()` as opposed to `urfave.NewBuilder()`) 

For reference, I've tried to express my thinking below
---
Architecture is a way of supporting some desirable characteristics of a software system, in a general sense we're managing risks related to its evolution, stability and performance. As an example, we know that the overlay network implementation may need to change, so we design an abstraction for it (mino). We know that ordering will need to change, so we do the same. We want to be able to run experiments, so we make it easy to have multiple swappable implementations.

In the CLI case, do we really expect to have use for multiple implementations? A `fake` CLI seems something that belongs in an integration test package, for instance. On the other hand, does this mean that if we want to change the CLI implementation, we have to go through each and every user of the `urfave` package, to point to a different package in the dependency injection system ? That seems cumbersome and 'urfave' isn't quite explicit as a package name.

Ultimately, it's the CLI so I don't expect there'll be too many places making use of this, and this design might help change the urfave-based CLI programs one-at-a-time if need be. It makes sense.

cli/urfave/mod.go
pierluca: agreed.
I was merely referring to the practice of returning concretions and letting the client code decide on the interface that suits its needs. In this case, I agree that it doesn't make much sense.

pierluca (approved):
I've added a couple of comments, but LGTM as-is.

PR #167, opened 2020-12-16
Router: Accept node's address

gnarula:
Some routing protocols require the router to know the node's address in order to decide the route.

This PR updates `Router.New` to accept the node's address. The `tree` router isn't changed since it doesn't make use of this information.

### DIFF
--- a/mino/minogrpc/example_test.go
+++ b/mino/minogrpc/example_test.go
@@ -57,14 +57,14 @@ func ExampleRPC_Call() {
 }
 
 func ExampleRPC_Stream() {
-	mA, err := NewMinogrpc(ParseAddress("127.0.0.1", 0), tree.NewRouter(NewAddressFactory()))
+	mA, err := NewMinogrpc(ParseAddress("127.0.0.1", 20000), tree.NewRouter(NewAddressFactory()))
 	if err != nil {
 		panic("overlay A failed: " + err.Error())
 	}
 
 	rpcA := mino.MustCreateRPC(mA, "test", exampleHandler{}, exampleFactory{})
 
-	mB, err := NewMinogrpc(ParseAddress("127.0.0.1", 0), tree.NewRouter(NewAddressFactory()))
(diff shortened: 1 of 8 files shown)


mino/router/mod.go
```diff
@@ -67,7 +67,7 @@ type Router interface {
 	GetHandshakeFactory() HandshakeFactory
 
 	// New creates a new routing table that will forward packets to the players.
```
nkcr, line 69: We should then update the comment to mention what the second parameter refers to, or use named parameters:

`New(players mino.Players, me mino.Address)`

mino/minogrpc/example_test.go
```diff
@@ -57,14 +57,14 @@ func ExampleRPC_Call() {
 }
 
 func ExampleRPC_Stream() {
-	mA, err := NewMinogrpc(ParseAddress("127.0.0.1", 0), tree.NewRouter(NewAddressFactory()))
+	mA, err := NewMinogrpc(ParseAddress("127.0.0.1", 20000), tree.NewRouter(NewAddressFactory()))
```
nkcr, line 60: Is there a particular reason for changing that ? Using port 0 makes the system assign a random free port.

mino/router/tree/mod.go
```diff
@@ -59,7 +59,7 @@ func (r Router) GetHandshakeFactory() router.HandshakeFactory {
 
 // New implements router.Router. It creates the routing table for the node that
 // is booting the protocol. This node will be the root of the tree.
-func (r Router) New(players mino.Players) (router.RoutingTable, error) {
+func (r Router) New(players mino.Players, address mino.Address) (router.RoutingTable, error) {
```
nkcr, line 62: a detail, but I find that `me` would be a better name than `address`.

mino/minogrpc/example_test.go
gnarula: I think the reason this was added was to make it easier to debug this Example. It takes a few more breakpoints to figure out the port for `mA` and `mB` but this way it's consistently `20000` and `30000`.

nkcr (approved):

Issue #162, opened 2020-10-14
Minogrpc certificate: hostname


Gilthoniel:
@gnarula you may want to address this for the student project.

Basically its generating a certificate for `127.0.0.1` but it should use the address hostname/IP.

```go
func (o *overlay) makeCertificate() error {
	tmpl := &x509.Certificate{
		SerialNumber: big.NewInt(1),
		IPAddresses:  []net.IP{net.ParseIP("127.0.0.1")},
```

It should instead check if its an IP or a hostname and fill the appropriate field of the certificate.

PR #168, opened 2020-12-16
minogrpc: use node's IP while creating certificate

gnarula:
Required to exchange certificates when creating multiple nodes in a containerised environment with each container having a unique IP rather than `127.0.0.1`

### DIFF
--- a/mino/minogrpc/server.go
+++ b/mino/minogrpc/server.go
@@ -18,7 +18,6 @@ import (
 	"time"
 
 	"go.dedis.ch/dela"
-	"go.dedis.ch/dela/internal/testing/fake"
 	"go.dedis.ch/dela/mino"
 	"go.dedis.ch/dela/mino/minogrpc/certs"
 	"go.dedis.ch/dela/mino/minogrpc/ptypes"
@@ -406,7 +405,7 @@ func (o *overlayServer) Forward(ctx context.Context, p *ptypes.Packet) (*ptypes.
 type overlay struct {
 	closer      *sync.WaitGroup
 	context     serde.Context
-	myAddr      mino.Address
+	myAddr      session.Address
 	certs       certs.Storage
 	tokens      tokens.Holder
 	router      router.Router
@@ -424,10 +423,8 @@ type overlay struct {
 }
 
 func newOverlay(tmpl minoTemplate) (*overlay, error) {
-	myAddrBuf, err := tmpl.myAddr.MarshalText()
-	if err != nil {
-		return nil, xerrors.Errorf("failed to marshal address: %v", err)
-	}
+	// session.Address never returns an error
+	myAddrBuf, _ := tmpl.myAddr.MarshalText()
 
 	if tmpl.secret == nil || tmpl.public == nil {
 		priv, err := ecdsa.GenerateKey(tmpl.curve, tmpl.random)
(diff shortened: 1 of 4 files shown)


nkcr:
closes #162

mino/minogrpc/server.go
```diff
@@ -549,9 +550,35 @@ func (o *overlay) Join(addr, token string, certHash []byte) error {
 }
 
 func (o *overlay) makeCertificate() error {
+	// We take a substring from [1, N) because
+	// o.myAddrStr[0] == 'F' || o.myAddrStr[1] == 'O'
+	hostname, _, err := net.SplitHostPort(o.myAddrStr[1:])
+	if err != nil {
+		if _, ok := o.myAddr.(fake.Address); ok {
```
gnarula, line 557: I'm not really fond of this check specifically. Another alternative is to have fake.Address.String() return `127.0.0.1` but that might be confusing. Whatsay?

nkcr:
I think we can update the overlay to store a `session.Address` instead of a `mino.Address`. This would then allow us to just use `o.myAddr.GetHostname()`.

I made the change in local to check the feasibility and there isn't much to change, I can push them on this PR is you want.

gnarula:
> I think we can update the overlay to store a `session.Address` instead of a `mino.Address`. This would then allow us to just use `o.myAddr.GetHostname()`.

I thought it's `mino.Address` so that we may pass a `fake.Address` in tests? The actual implementation uses a `session.Address`  https://github.com/dedis/dela/blob/57974051932488cccba68e1da819f7288f2bbf5e/mino/minogrpc/mod.go#L149

Now that I think of it, myAddr would always point to an IP and not a hostname since `NewMinogrpc` accepts a `net.Addr` where `net.Addr.String()` points to an `IPv4:IPv6:port` string.

nkcr:
> I thought it's `mino.Address` so that we may pass a `fake.Address` in tests? The actual implementation uses a `session.Address`

In the tests there is only one case where we really need to passe a fake address, it is to check `func newOverlay(tmpl minoTemplate) (*overlay, error)` and the case where `tmpl.myAddr.MarshalText()` returns an error. But if we change the template to contain a session.Address, we can be sure that `MarshalText()` will never return an error. Then we can remove this check:

```
func newOverlay(tmpl minoTemplate) (*overlay, error) {
	// the session address never returns an error
	myAddrBuf, _ := tmpl.myAddr.MarshalText()
	...
```

and in the tests use `session.Address{}` when needed, instead of `fake.NewAddress(0)`.

mino/minogrpc/server.go
```diff
@@ -549,9 +547,29 @@ func (o *overlay) Join(addr, token string, certHash []byte) error {
 }
 
 func (o *overlay) makeCertificate() error {
+	hostname, err := o.myAddr.GetHostname()
+	if err != nil {
+		return xerrors.Errorf("error parsing node's hostname: %v", err)
+	}
+
+	ipAddr := net.ParseIP(hostname)
```
nkcr, line 555: we can simplify the whole by just calling once `ipAddrs, err := net.LookupIP(hostname)`

mino/minogrpc/server.go
```diff
@@ -549,9 +547,18 @@ func (o *overlay) Join(addr, token string, certHash []byte) error {
...
 func (o *overlay) makeCertificate() error {
+	hostname, err := o.myAddr.GetHostname()
+	if err != nil {
+		return xerrors.Errorf("error retrieving hostname: %v", err)
+	}
+
+	ipAddrs, err := net.LookupIP(hostname)
+	if err != nil {
+		return xerrors.Errorf("error resolving IP: %v", err)
+	}
```
nkcr, line 558: Just a new line here and LGTM 👍

nkcr (approved):

nkcr (approved):

PR #169, opened 2020-12-17
minogrpc: handle multiple messages in stream

gnarula:
minogrpc's GRPC Stream handler calls `endpoint.Handler.Stream` once and then waits for the stream to close. This change loops over `endpoint.Handler.Stream` indefinitely on a separate go routine allowing multiple messages to be processed in the same `Session` for a child node.

An example use case would be in a simnet simulation, where a session/routing tables are established in the [Before](https://github.com/dedis/student20_rabyt/blob/82cf48b827082d3a3194eba10e998c01c0fcc7ef/simulation/main.go#L20) step and more messages would be sent in the [Execute](https://github.com/dedis/student20_rabyt/blob/82cf48b827082d3a3194eba10e998c01c0fcc7ef/simulation/main.go#L66) step. The sender/receiver instances are cached on the orchestrator. Refer to this [link](https://github.com/dedis/student20_rabyt/blob/82cf48b827082d3a3194eba10e998c01c0fcc7ef/controller/actions.go#L128) for more info.

This is likely leaking a go-routine - though I'm not entirely sure about it. Consider a case where the go-routine is blocked on `endpoint.Handler.Stream` and the stream go-routine returns after receiving from `stream.Context().Done()`. Perhaps `endpoint.Handler.Stream` should be modified to accept a context to indicate the parent is going to exit?

### DIFF
--- a/mino/minogrpc/server.go
+++ b/mino/minogrpc/server.go
@@ -309,14 +309,25 @@ func (o *overlayServer) Stream(stream ptypes.Overlay_StreamServer) error {
 		return xerrors.Errorf("failed to send header: %v", err)
 	}
 
-	err = endpoint.Handler.Stream(sess, sess)
-	if err != nil {
-		return xerrors.Errorf("handler failed to process: %v", err)
+	errs := make(chan error)
+	go func() {
+		for {
+			err = endpoint.Handler.Stream(sess, sess)
+			if err != nil {
+				errs <-err
+				return
+			}
+		}
+	}()
+
+	for {
+		select {
+		case err := <-errs:
+			return xerrors.Errorf("handler failed to process: %v", err)
+		case <-stream.Context().Done():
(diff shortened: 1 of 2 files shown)


nkcr:
> and more messages would be sent in the Execute step

Then it should be the handler's (`mino.Handler`) responsibility to not end, but keep looping. You can have a handler that loops forever, like this is the case for [cosi](https://github.com/dedis/dela/blob/master/cosi/threshold/handler.go#L36).

gnarula:
> > and more messages would be sent in the Execute step
> 
> Then it should be the handler's (`mino.Handler`) responsibility to not end, but keep looping. You can have a handler that loops forever, like this is the case for [cosi](https://github.com/dedis/dela/blob/master/cosi/threshold/handler.go#L36).

That solves it! Thanks :)

PR #172, opened 2021-01-04
Adds a pool controller

nkcr:
This PR add a controller to the pool, which allows one to send transactions to the pool.
The new command is of form:

```
memcoin [--config CONFIG] pool add --key KEY [{--args KEY --args VALUE} ...] [--nonce NONCE]

### DIFF
--- a/core/txn/pool/controller/action.go
+++ b/core/txn/pool/controller/action.go
@@ -1,6 +1,8 @@
 package controller
 
 import (
+	"sync"
+
 	"go.dedis.ch/dela/crypto"
 
 	"go.dedis.ch/dela/cli/node"
@@ -22,11 +24,16 @@ var getManager = func(signer crypto.Signer, s signed.Client) txn.Manager {
 //
 // - implements node.ActionTemplate
 type addAction struct {
+	sync.Mutex
+
 	client *client
 }
 
 // Execute implements node.ActionTemplate
-func (a addAction) Execute(ctx node.Context) error {
+func (a *addAction) Execute(ctx node.Context) error {
+	a.Lock()
+	defer a.Unlock()
+
 	var p pool.Pool
 	err := ctx.Injector.Resolve(&p)
 	if err != nil {
--- a/core/txn/pool/controller/action.go
+++ b/core/txn/pool/controller/action.go
@@ -1,0 +1,84 @@
+package controller
+
+import (
+	"go.dedis.ch/dela/crypto"
+
+	"go.dedis.ch/dela/cli/node"
+	"go.dedis.ch/dela/core/txn"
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: core/txn/pool/controller/mod.go [hunk 1]
```diff
@@ -0,0 +1,67 @@
...
+func (m miniController) OnStart(flags cli.Flags, inj node.Injector) error {
+	return nil
+}
+
+// OnStop implements node.Initializer
+func (miniController) OnStop(inj node.Injector) error {
+	return nil
+}
+
+// client return monotically inscreasing nonce
```
gnarula, line 55: s/inscreasing/increasing

core/txn/pool/controller/mod.go
```diff
@@ -0,0 +1,67 @@
...
+//
+// - implements signed.Client
+type client struct {
+	nonce uint64
+}
+
+// GetNonce implements signed.Client
+func (c *client) GetNonce(access.Identity) (uint64, error) {
+	res := c.nonce
+	c.nonce++
```
gnarula, line 65: Is there a race possible in this? i.e. do we expect the client to be run concurrently? Similar doubt for `TransactionManager.Make`

core/txn/pool/controller/action.go
```diff
@@ -0,0 +1,84 @@
...
+		return xerrors.Errorf("failed to load signer: %v", err)
+	}
+
+	signer, err := bls.NewSignerFromBytes(signerdata)
+	if err != nil {
+		return xerrors.Errorf("failed to unmarshal signer: %v", err)
+	}
+
+	nonce := ctx.Flags.Int("nonce")
+	if nonce != -1 {
```
gnarula, line 62: I'd suggest we raise an error for nonce values < -1 or add a `UIntFlag` to avoid a huge nonce in case the user enters a negative value

nkcr: the system will refuse a nonce that is not strictly following the last one. Ie. if the last nonce was 9, the only valid possibility is to provide nonce 10, otherwise it returns an error.

core/txn/pool/controller/mod.go
gnarula: Tried a small experiment with https://github.com/dedis/dela/commit/4a0ed16f988b7a6dc2bcecf45f2ce801d8e2a7f2 and it is indeed possible.

nkcr: ok, nice catch!

gnarula (approved):
Looks good!

pierluca (commented):
Minor suggestions, but LGTM overalls 👍

pierluca (approved):

PR #173, opened 2021-01-04
Adds the access contract

nkcr:
Adds an access contract that allows an authorized identity to grant access on the other smart contracts. Each node must locally grant access to the identity that can use this smart contract, with the cli, in the form:

```
memcoin [--config /tmp/node1] access add [--identity IDENTITY ...]
```

As this access is critical and can be different on each node, it is kept in a separate storage on each server and not part of the blockchain.

### DIFF
--- a/contracts/access/controller/action.go
+++ b/contracts/access/controller/action.go
@@ -1,3 +1,8 @@
+// This file implements the action of the controller.
+//
+// Documentation Last Review: 02.02.2021
+//
+
 package controller
 
 import (
@@ -18,7 +23,7 @@ import (
 type addAction struct{}
 
 // Execute implements node.ActionTemplate. It reads the list of identities and
-// update the access.
+// updates the access.
 func (a addAction) Execute(ctx node.Context) error {
 	var exec *native.Service
 	err := ctx.Injector.Resolve(&exec)
@@ -39,20 +44,9 @@ func (a addAction) Execute(ctx node.Context) error {
 	}
 
 	idsStr := ctx.Flags.StringSlice("identity")
-	identities := make([]access.Identity, len(idsStr))
-
-	for i, id := range idsStr {
-		idBuf, err := base64.StdEncoding.DecodeString(id)
-		if err != nil {
-			return xerrors.Errorf("failed to decode pub key '%s': %v", id, err)
-		}
-
-		pk, err := bls.NewPublicKey(idBuf)
-		if err != nil {
-			return xerrors.Errorf("failed to unmarshal identity '%s': %v", id, err)
-		}
-
-		identities[i] = pk
+	identities, err := parseIdentities(idsStr)
+	if err != nil {
+		return xerrors.Errorf("failed to parse identities: %v", err)
 	}
 
 	err = asrv.Grant(accessStore, accessContract.NewCreds(aKey[:]), identities...)
@@ -63,4 +57,24 @@ func (a addAction) Execute(ctx node.Context) error {
 	dela.Logger.Info().Msgf("access granted to %v", identities)
 
 	return nil
+}
+
+func parseIdentities(idsStr []string) ([]access.Identity, error) {
+	identities := make([]access.Identity, len(idsStr))
+
+	for i, id := range idsStr {
+		idBuf, err := base64.StdEncoding.DecodeString(id)
+		if err != nil {
+			return nil, xerrors.Errorf("failed to decode pub key '%s': %v", id, err)
+		}
+
+		pk, err := bls.NewPublicKey(idBuf)
+		if err != nil {
+			return nil, xerrors.Errorf("failed to unmarshal identity '%s': %v", id, err)
+		}
+
+		identities[i] = pk
+	}
+
+	return identities, nil
 }
--- a/contracts/access/controller/action.go
+++ b/contracts/access/controller/action.go
@@ -1,0 +1,66 @@
+package controller
+
+import (
+	"encoding/base64"
+
+	"go.dedis.ch/dela"
+	"go.dedis.ch/dela/cli/node"
+	accessContract "go.dedis.ch/dela/contracts/access"
+	"go.dedis.ch/dela/core/access"
+	"go.dedis.ch/dela/core/execution/native"
+	"go.dedis.ch/dela/crypto/bls"
+	"golang.org/x/xerrors"
+)
+
+// addAction is an action to add one or more identities.
+//
+// - implements node.ActionTemplate
+type addAction struct{}
+
+// Execute implements node.ActionTemplate. It reads the list of identities and
+// update the access.
+func (a addAction) Execute(ctx node.Context) error {
+	var exec *native.Service
+	err := ctx.Injector.Resolve(&exec)
+	if err != nil {
+		return xerrors.Errorf("failed to resolve native service: %v", err)
+	}
+
+	var asrv access.Service
+	err = ctx.Injector.Resolve(&asrv)
+	if err != nil {
+		return xerrors.Errorf("failed to resolve access service: %v", err)
+	}
+
+	var accessStore accessStore
+	err = ctx.Injector.Resolve(&accessStore)
+	if err != nil {
+		return xerrors.Errorf("failed to resolve access store: %v", err)
+	}
+
+	idsStr := ctx.Flags.StringSlice("identity")
+	identities := make([]access.Identity, len(idsStr))
+
+	for i, id := range idsStr {
+		idBuf, err := base64.StdEncoding.DecodeString(id)
+		if err != nil {
+			return xerrors.Errorf("failed to decode pub key '%s': %v", id, err)
+		}
+
+		pk, err := bls.NewPublicKey(idBuf)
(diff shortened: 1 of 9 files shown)


contracts/access/controller/jstore.go
```diff
@@ -0,0 +1,102 @@
...
+	return s.data[string(key)], nil
+}
+
+func (s *jstore) saveFile() error {
+	buf, err := s.ctx.Marshal(s.data)
+	if err != nil {
+		return xerrors.Errorf("failed to marshal data: %v", err)
+	}
+
+	err = ioutil.WriteFile(s.path, buf, os.ModePerm)
```
gnarula, line 91: Perhaps the permissions here can be more restrictive? I'm thinking about `0644` (which would essentially be a `0622` after a default umask of `0022`) or `0600`.

contracts/access/mod.go
```diff
@@ -0,0 +1,171 @@
+// Package access implements a native contract to handle access. It allows an
+// authorized identity to add access as an {ID, CONTRACT, COMMAND, IDENTITIES}
+// quadruplet.
+//
+// ID is the credential identifier. This identifier is generally defined at the
+// contract's creation.
+// CONTRACT is the contract name.
+// COMMAND specifies the command to grant access to on the contract.
+// IDENTITIES is a list of base64 encoded bls public keys, separated by comas.
```
gnarula, line 9:
```suggestion
// IDENTITIES is a list of standard base64 encoded bls public keys, separated by comas.
```

contracts/access/mod.go
```diff
@@ -0,0 +1,171 @@
...
+
+// Command defines a command for the value contract
+type Command string
+
+const (
+	// CmdSet defines the command to grant access
+	CmdSet Command = "GRANT"
+)
+
+// NewCreds creates new credentials for a value contract execution.
```
gnarula, line 59:
```suggestion
// NewCreds creates new credentials for an access contract execution.
```

contracts/access/mod.go
```diff
@@ -0,0 +1,171 @@
...
+func NewCreds(id []byte) access.Credential {
+	return access.NewContractCreds(id, ContractName, "all")
```
gnarula, line 61: does `all` have a special meaning here? IIUC, is it a credential for all commands in the contract?

contracts/access/mod.go
```diff
@@ -0,0 +1,171 @@
...
+		accessKey: aKey,
+		store:     store,
+	}
+}
+
+// Execute implements native.Contract
+func (c Contract) Execute(snap store.Snapshot, step execution.Step) error {
+	creds := NewCreds(c.accessKey)
+
+	err := c.access.Match(c.store, creds, step.Current.GetIdentity())
```
gnarula, line 96: Continuing from above, is this check for ensuring the transaction executor's identity is allowed to execute a command in the access contract?

I'm not quite clear on where the initial permissions for the access contract are defined

contracts/access/mod.go
nkcr: There is no special meaning there, one can define multiple credentials with different names. This would be useful if we wanted to have different credential for different commands on the contract. Here we use only one credential that is named "all" and grants access to everything.

contracts/access/mod.go
nkcr: In `NewContract` we provide the `credential.Access` service, which is the one we use there. In the case of this contract, this service is not the regular access service that the blockchain uses, but a local one to the node, which is stored using `jstore.go`. You can see it in `controller/mod.go:63` when we create the contract. A "regular" contract would use to injector to retrieve the access service, but there we are creating a new local one, which can be updated locally with the cli. 

And yes, this check is to ensure that the identity has been (locally in this case) allowed to updated the access contract.

gnarula: > You can see it in `controller/mod.go:63` when we create the contract. A "regular" contract would use to injector to retrieve the access service, but there we are creating a new local one, which can be updated locally with the cli.

You probably meant resolving the access store there. So if I understand correctly, the invocation of the contract would result in all the local `jstores` being updated since the contract is executed and successful state changes are applied at all the nodes?

Btw, I couldn't an implementation of access.Service (probably `darc`?) being injected anywhere but the command assumes it's available to the injector. Am I overlooking something?

nkcr: In Dela there are now two different access services:

- the DARC one, which is the one used to control access to all smart contract, except the "access smart contract". The DARC access is the same on every node. It is kept in sync because we update it with transactions that we perform on the "access smart contract". The "access smart contract" is responsible for updating the DARC access. The DARC access is created [there](https://github.com/dedis/dela/blob/e1cc707548c7ab4d7cea0782ba76d2f24d8e2079/core/ordering/cosipbft/controller/mod.go#L122). It is injected in PR #174.

- the "local access", which is only used to control access on the "access smart contract". This access is local to every node and can be changed by running a CLI command locally. This access is stored in a json file (jstore.go).

In summary: the access smart contract is responsible for **updating the DARC access**. Access on the access smart contract is **not controlled by the DARC**, as it is the case for all other smart contracts, but by the **local access**, which we can update only with a local CLI command.

gnarula: That clears it! Thanks 👍

gnarula (approved):

contracts/access/mod.go
```diff
@@ -0,0 +1,172 @@
+// Package access implements a native contract to handle access. It allows an
```
pierluca, line 1: Should we keep adding a `// Documentation Last Review: 01.02.2021` as @Gilthoniel had suggested?
It might be useful :)

... same comment in other files too :)

contracts/access/mod.go
```diff
@@ -0,0 +1,172 @@
...
+type Command string
+
+const (
+	// CmdSet defines the command to grant access
+	CmdSet Command = "GRANT"
+)
+
+// NewCreds creates new credentials for an access contract execution.
+func NewCreds(id []byte) access.Credential {
+	return access.NewContractCreds(id, ContractName, "all")
```
pierluca, line 62: same remark as the other PR, especially given @gnarula had the same question.
It might be good to put that "all" in a named constant for clarity.

contracts/access/mod.go
```diff
@@ -0,0 +1,172 @@
...
+// grant perform the GRANT command
+func (c Contract) grant(snap store.Snapshot, step execution.Step) error {
+	idHex := step.Current.GetArg(GrantIDArg)
+	if len(idHex) == 0 {
+		return xerrors.Errorf("'%s' not found in tx arg", GrantIDArg)
+	}
+
+	id, err := hex.DecodeString(string(idHex))
+	if err != nil {
+		return xerrors.Errorf("failed to decode id: %v", err)
```
pierluca, line 129: should the error point out `in tx arg` ?

contracts/access/controller/jstore.go
```diff
@@ -0,0 +1,102 @@
...
+		buf, err := ioutil.ReadFile(path)
+		if err != nil {
+			return nil, xerrors.Errorf("failed to read file '%s': %v", path, err)
+		}
+
+		err = ctx.Unmarshal(buf, &data)
+		if err != nil {
+			return nil, xerrors.Errorf("failed to read json: %v", err)
+		}
+	}
```
pierluca, line 35: There's a subtle issue with this.
If `fileExist(path)`, we might identify issues at startup (corruption, lacking access rights, etc.
If it doesn't, we'll only find out much later that something's wrong (on a call to `saveFile`, most likely).

Would it make sense to create the file immediately, possibly even save after read, to make sure that all foreseeable problems have been detected at startup? (e.g. no write access, parent folder doesn't exist, etc.)

contracts/access/controller/action.go
```diff
@@ -0,0 +1,66 @@
...
+			return xerrors.Errorf("failed to decode pub key '%s': %v", id, err)
+		}
+
+		pk, err := bls.NewPublicKey(idBuf)
+		if err != nil {
+			return xerrors.Errorf("failed to unmarshal identity '%s': %v", id, err)
+		}
+
+		identities[i] = pk
+	}
```
pierluca, line 56: this could be extracted in a separate function to improve readability.

contracts/access/controller/action.go
```diff
@@ -0,0 +1,66 @@
...
+	"golang.org/x/xerrors"
+)
+
+// addAction is an action to add one or more identities.
+//
+// - implements node.ActionTemplate
+type addAction struct{}
+
+// Execute implements node.ActionTemplate. It reads the list of identities and
+// update the access.
```
pierluca, line 21:
```suggestion
// updates the access.
```

pierluca (commented):
As usual, I would strongly suggest splitting up the test cases :) 
Other than that, a few minor things here and there, but LGTM overalls 👍

contracts/access/controller/jstore.go
nkcr: nice catch, makes sense!

pierluca (approved):
LGTM !

PR #174, opened 2021-01-05
Adds the value contract

nkcr:
This PR adds a simple value contract that allows an authorized identity to perform CRUD operations on the storage.


gnarula (approved):

contracts/value/mod.go
```diff
@@ -0,0 +1,245 @@
...
+
+	// CmdRead defines a command to read a value
+	CmdRead Command = "READ"
+
+	// CmdDelete defines a command to delete a value
+	CmdDelete Command = "DELETE"
+
+	// CmdDisplay defines a command to display all values set (and not deleted)
+	// so far.
+	CmdDisplay Command = "DISPLAY"
```
pierluca, line 60: wouldn't this be a sort of `LIST` then?
I feel `DISPLAY` doesn't entail that we're fetching all values.

contracts/value/mod.go
```diff
@@ -0,0 +1,245 @@
...
+
+// Command defines a type of command for the value contract
+type Command string
+
+const (
+	// CmdSet defines the command to set a value
+	CmdSet Command = "SET"
+
+	// CmdRead defines a command to read a value
+	CmdRead Command = "READ"
```
pierluca, line 53: Would it make sense to adopt either `GET`/`SET` or `READ`/`WRITE` for consistency?

I have a personal preference for `READ`/`WRITE`, in this context, but anything consistent will do :)

contracts/value/mod.go
```diff
@@ -0,0 +1,245 @@
...
+
+	// CmdDisplay defines a command to display all values set (and not deleted)
+	// so far.
+	CmdDisplay Command = "DISPLAY"
+)
+
+// NewCreds creates new credentials for a value contract execution. We might
+// want to use in the future a separate credential for each command.
+func NewCreds(id []byte) access.Credential {
+	return access.NewContractCreds(id, ContractName, "all")
```
pierluca, line 66: what's `"all"` ? I'd consider putting that in a variable just to document the intention.

contracts/value/mod.go
```diff
@@ -0,0 +1,245 @@
...
+	access access.Service
+
+	// accessKey is the identity's key allowed to use this smart contract
+	accessKey []byte
+
+	// cmd provides the commands executions
+	cmd commands
+
+	// printer is the output used by the READ and DISPLAY commands
+	printer io.Writer
```
pierluca, line 92: This documentation isn't very clear to me.
I've tried to rewrite it, but I'm not sure I understood correctly

```suggestion
	// access is the access control service managing this smart contract
	access access.Service

	// accessKey is the identity's key allowed to use this smart contract (????)
	accessKey []byte

	// cmd provides the commands that can be executed by this smart contract
	cmd commands

	// printer is the output used by the READ and DISPLAY commands (????)
	printer io.Writer
```

Why would we "print" a value rather than return it, when executing this smart contract?
That confuses me a bit as well

contracts/value/mod.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	err := snap.Set(key, value)
+	if err != nil {
+		return xerrors.Errorf("failed to set value: %v", err)
+	}
+
+	c.index[string(key)] = struct{}{}
+
+	dela.Logger.Info().Str("contract", "value").Msgf("setting %s=%s", key, value)
```
pierluca, line 178: If I understand their purpose correctly, this looks like `"contract"` and `"value"` could be defined as constants somewhere. Shouldn't we use `ContractName` ?

contracts/value/mod.go
```diff
@@ -0,0 +1,245 @@
+// Package value implements a simple native contract that can store, delete, and
```
pierluca, line 1: I'm not sure I understand correctly, but if I do: 

1) I'm a bit confused by the fact that reading data is only useful for logging.

2) It's not possible to list (`display`) values in the store/snapshot: only the ones that have been written within the current execution of the Dela node will be listable. If I understand correctly, it means that two nodes (with perfect uptime) joining at different points in time the ledger might behave differently for the `display` command... do I get this right? If so, that seems very wrong.

contracts/value/mod_test.go
```diff
@@ -0,0 +1,235 @@
...
+	snap.Set([]byte("key1"), []byte("value1"))
+	snap.Set([]byte("key2"), []byte("value2"))
+
+	err := cmd.display(snap)
+	require.NoError(t, err)
+
+	require.Equal(t, "key1=value1,key2=value2", buf.String())
+
+	err = cmd.display(fake.NewBadSnapshot())
+	// we can't assume an order from the map
```
pierluca, line 156: how does this comment relate to the assert?

contracts/value/mod_test.go
```diff
@@ -0,0 +1,235 @@
...
+	require.Nil(t, res)
+
+	_, found := contract.index["dummy"]
+	require.False(t, found)
+}
+
+func TestCommand_Display(t *testing.T) {
+	contract := NewContract([]byte{}, fakeAccess{})
+	contract.index["key1"] = struct{}{}
+	contract.index["key2"] = struct{}{}
```
pierluca, line 137: what guarantees that this would happen on startup?

pierluca (changes requested):
A few questions here and there.

General remark on tests: split out the different cases in different functions. This ensures much better independence and reporting when something actually goes wrong.

contracts/value/mod.go
nkcr: 1. When you send a transaction, there is no notion of "returned value". It works like the following: A client can only send a transaction to the pool, which is a simple operation that returns once the transaction has been submitted. From there, the client can only wait for the leader to include that transaction in a block and, if the block is accepted, that transaction will be executed on each node and update each node's state. The only mean we have to "get  a value" (at least right now) is to display that value upon the transaction's execution.

2. A snapshot only provides Get/Set/Delete operations. So you can only get the values whose keys are known to you. 

> do I get this right?

It depends on the joining mechanism. If we require a joining node to only load the current database, yes, it will not display the same thing as the other nodes. If we make a new node replay the chain then no, it will have the exact same state (`c.index`) as the other nodes, displaying the same result.

contracts/value/mod.go
nkcr: I'm sold with `READ` / `WRITE` !

contracts/value/mod.go
nkcr: that's the credential name, we could use different credentials for each action, ie. one for READ, WRITE, DISPLAY, so that's where we would use a different name. There we only have one global credential call "all".

contracts/value/mod_test.go
nkcr: I can't check for the exact error string, because it could either say that key1 is missing, or key2.

contracts/value/mod.go
pierluca: Re. point (1), I believe I understand the notion of transaction without a return value.
However, my thinking was with respect to contract interactions.  Can contracts call upon other contracts? I assume not yet, but do we want to support that in the future? If so, the output of one might inform the input of another, correct?

Or are you saying that this is so far out into the future, and requiring so many changes, that this would be the least of our worries? 😅

contracts/value/mod_test.go
pierluca: ahhhhh, sorry, now I get it:
```suggestion
	// check for a generic error, exact key depends on undetermined map order
```

contracts/value/mod_test.go
nkcr: none. This is to manually fill the index, so we can later send the display command.

contracts/value/mod.go
nkcr: > Or are you saying that this is so far out into the future, and requiring so many changes, that this would be the least of our worries? 😅

Yes. The interface is not ready for that.

pierluca (approved):

PR #175, opened 2021-01-06
Implements the BLS crypto CLI

nkcr:
This CLI is needed is we want to create identities/signatures that can send transactions to the pool. This is where having an independent CLI implementation was important (#165).


crypto/bls/command/mod.go
```diff
@@ -0,0 +1,58 @@
...
+
+	read := signer.SetSubCommand("read")
+	read.SetDescription("read a signer")
+	read.SetFlags(cli.StringFlag{
+		Name:     "path",
+		Usage:    "path to the signer's file",
+		Required: true,
+	}, cli.StringFlag{
+		Name:     "format",
+		Usage:    "output format: [PUBKEY | BASE64 | BIN_PUBKEY]",
```
gnarula, line 53:
```suggestion
		Usage:    "output format: [PUBKEY | BASE64 | BASE64_PUBKEY]",
```

gnarula (changes requested):

cli/crypto/mod.go
```diff
@@ -0,0 +1,37 @@
...
+	"io"
+	"os"
+
+	"go.dedis.ch/dela/cli"
+	"go.dedis.ch/dela/cli/ucli"
+	bls "go.dedis.ch/dela/crypto/bls/command"
+)
+
+var builder cli.Builder = ucli.NewBuilder("crypto", nil)
+var printer io.Writer = os.Stdout
```
pierluca, line 16: it seems we only use it for an error. Shouldn't it be os.Stderr ?

cli/crypto/mod_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"go.dedis.ch/dela/cli"
+)
+
+func TestMain(t *testing.T) {
+	builder = &fakeBuilder{}
+	main()
```
pierluca, line 14: what is the point of this code?
No asserts are being made about it and it seems irrelevant to the rest of the test.

pierluca: if the idea is to test the 'happy path' , consider splitting it out in a separate test asserting that nothing is sent on the printer.

crypto/bls/command/action.go
```diff
@@ -0,0 +1,118 @@
...
+
+	return nil
+}
+
+func saveToFile(path string, force bool, data []byte) error {
+	existMsg := "file '%s' already exist, use --force if you want to overwrite"
+
+	exist := fileExist(path)
+	if !force && exist {
+		return xerrors.Errorf(existMsg, path)
```
pierluca, line 98: why not put the existMsg directly here?

crypto/bls/command/action.go
```diff
@@ -0,0 +1,118 @@
...
+	fmt.Fprintln(a.printer, string(out))
+
+	return nil
+}
+
+func saveToFile(path string, force bool, data []byte) error {
+	existMsg := "file '%s' already exist, use --force if you want to overwrite"
+
+	exist := fileExist(path)
+	if !force && exist {
```
pierluca, line 97:
```suggestion
	if !force && fileExist(path) {
```

crypto/bls/command/action.go
```diff
@@ -0,0 +1,118 @@
...
+		return xerrors.Errorf(existMsg, path)
+	}
+
+	ioutil.WriteFile(path, data, os.ModePerm)
```
pierluca, line 101: shouldn't we process & report the error from WriteFile ?

crypto/bls/command/action_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+func TestGetPUBKEY(t *testing.T) {
+	_, err := getPubkey(nil)
+	require.EqualError(t, err, "failed to unmarshal signer: while unmarshaling scalar: UnmarshalBinary: wrong size buffer")
+
+	buf, err := bls.NewSigner().MarshalBinary()
+	require.NoError(t, err)
+
+	_, err = getPubkey(buf)
+	require.NoError(t, err)
```
pierluca, line 115: I'd split out this (failure) test case from the previous (successful) case.

crypto/bls/command/action_test.go
```diff
@@ -0,0 +1,151 @@
+package command
```
pierluca, line 1: In general, each single test function here is testing A LOT of different cases.

It'd be best to split out the test cases in different functions.
That way when a function fails, we immediately know the case that's affected.

pierluca (changes requested):
There seems to be a few tiny issues.
I'd also consider refactoring some tests.
Overalls it's quite clear though :)

crypto/bls/command/action.go
nkcr: nice catch!

pierluca (approved):

PR #176, opened 2021-01-11
Updates the http proxy

nkcr:
This update adds a command to manually start the http proxy. It also adds the GetAddr() function to the interface.


mino/proxy/http/controller/action.go
```diff
@@ -0,0 +1,28 @@
...
+// server.
+func (a startAction) Execute(ctx node.Context) error {
+	addr := ctx.Flags.String("clientaddr")
+
+	proxyhttp := http.NewHTTP(addr)
+
+	ctx.Injector.Inject(proxyhttp)
+
+	go proxyhttp.Listen()
+	time.Sleep(time.Second)
```
pierluca, line 23: why do we need a `time.Sleep` here? 
It looks a bit wrong.
Are we waiting for Listen() to be operational? If so, isn't there a status we can check?

mino/proxy/http/controller/action_test.go
```diff
@@ -0,0 +1,27 @@
...
+	ctx := node.Context{
+		Injector: node.NewInjector(),
+		Flags:    flags,
+		Out:      out,
+	}
+
+	action := startAction{}
+	action.Execute(ctx)
+
+	require.Equal(t, "started proxy server on 127.0.0.1:3000", out.String())
```
pierluca, line 26: we're only asserting on logging, shouldn't we assert that the clientaddr now has a HTTP proxy running?

mino/proxy/http/mod_test.go
```diff
@@ -27,6 +29,47 @@ func TestHTTP_Listen(t *testing.T) {
 	require.Equal(t, "hello", string(output))
 }
 
+func TestHTPP_Listen_EmptyAddr(t *testing.T) {
+	// in this case it will use a random free port
```
pierluca, line 33: there are no asserts in this test. Shouldn't we be asserting something?

mino/proxy/http/mod_test.go
```diff
@@ -27,6 +29,47 @@ func TestHTTP_Listen(t *testing.T) {
...
+	proxy := NewHTTP("")
+
+	go proxy.Listen()
+	time.Sleep(200 * time.Millisecond)
+
+	defer proxy.Stop()
```
pierluca, line 39: no need for defer, right?
or at least, not after `time.Sleep` ?

mino/proxy/http/mod_test.go
```diff
@@ -27,6 +29,47 @@ func TestHTTP_Listen(t *testing.T) {
...
+
+func TestHTPP_Listen_BadAddr(t *testing.T) {
+	proxy := NewHTTP("bad://xx")
+
+	out := new(bytes.Buffer)
+	proxy.logger = zerolog.New(zerolog.ConsoleWriter{Out: out})
+
+	go func() {
+		defer func() {
+			recover()
```
pierluca, line 50: we can verify the `recover()` output to check that a panic was raised.

pierluca (changes requested):
A few odd things in the timing / tests , worth looking into.
Other than that, seems straightforward :)

mino/proxy/http/controller/action.go
nkcr: With net/http there is no nice way that I know to do that. We could also poll the GetAddr() until we have an address. For now 1 sec is the reasonable amount of time we choose to check if anything panics.

mino/proxy/http/mod.go
```diff
@@ -97,6 +105,11 @@ func (h HTTP) Stop() {
 	h.quit <- struct{}{}
 }
 
+// GetAddr implements proxy.Proxy.
+func (h HTTP) GetAddr() net.Addr {
+	return h.ln.Addr()
```
pierluca, line 110: calling this before `Listen()` could result in `h.ln` not being initialized.
Should we guard against that case ?

mino/proxy/http/controller/action.go
pierluca: My concern is that `time.Sleep` is brittle. If there's a temporary freeze, or anything happens in the system, the next line will fail.

We could do something like this, though, no? 

```suggestion
        for proxyhttp.GetAddr() == nil {
	    time.Sleep(time.Second)
	}
```

Of course we'd then need an overall timeout to make sure that we don't go in an infinite loop if for whatever reason the Listen() fails

mino/proxy/http/controller/action.go
```diff
@@ -0,0 +1,28 @@
...
+
+	fmt.Fprintf(ctx.Out, "started proxy server on %s", proxyhttp.GetAddr().String())
```
pierluca, line 25: Listen can fail, here we assume that it's always going to be successful.
Shouldn't we ask of Listen() to return an error code?

nkcr: Let's add that later on a new PR if we feel it's needed. I don't want to rush into updating the proxy interface without taking the time to properly update the tests.

pierluca: Works for me.
You might want to add a comment, so that whoever revisits this code knows of that implicit assumption though.
It'll save us time when (eventually) the problem shows up :)

pierluca (approved):

pierluca:
I've approved, but are you sure that `GetAddr()` returns nil rather than panic, if the the server isn't started?

nkcr:
> I've approved, but are you sure that GetAddr() returns nil rather than panic, if the the server isn't started?

That wasn't the case, but I've changed the implementation to behave like so.

PR #181, opened 2021-02-22
Adds instructions and minor fixes to use Memcoin

### LINKED ISSUE #177: Ordering: rotate leader
Right now when a round expires, nodes choose the leader + 1, without rotating:

```
	param := ViewParam{
		From:   addr,
		ID:     lastID,
		Leader: m.round.leader + 1,
	}
```

we should apply a modulo based on the number of nodes.

### LINKED ISSUE #178: Ordering: remove unvalid transaction
A transaction that fails to be executed (ie. when its execution returns an error) should be removed from the pool. Otherwise we might get an infinite loop, because the node will always re-try to execute it.

nkcr:
Fixes #178.  
Fixes #177.


docs/memcoin.md
```diff
@@ -0,0 +1,60 @@
...
+memcoin --config /tmp/node1 access add \
+    --identity $(crypto bls signer read --path private.key --format BASE64_PUBKEY)
+memcoin --config /tmp/node2 access add \
+    --identity $(crypto bls signer read --path private.key --format BASE64_PUBKEY)
+memcoin --config /tmp/node3 access add \
+    --identity $(crypto bls signer read --path private.key --format BASE64_PUBKEY)
+
+# Update the access contract to allow us to use the value contract
+memcoin --config /tmp/node1 pool add\
+    --key private.key\
```
gnarula, line 39: This should be the absolute path to `private.key`. I got the following error when I tried the relative path as in the example:

```
command error: failed to get signer: failed to load signer: while opening file: open private.key: no such file or directory
```

gnarula (changes requested):
Looks good but I've a minor comment in the documentation. Would be nice to support relative path instead of requiring an absolute path or even both but maybe that's a separate PR on its own.

docs/memcoin.md
nkcr: Thank you for trying it out. It is possible to use an absolute, and the path is relative to where the node has been started. I added a note about it.

gnarula (approved):

PR #182, opened 2021-03-09
add benchmark for increment contract in solidity

gnarula:
Adds an evm execution type that tries to run the `increment` solidity contract.

* Compile the test using `go test -c ./...` in the `benchmark` directory
* Run the benchmark in the `evm` directory `/../benchmark/benchmark.test -test.bench=.`

Here's a sample run

```
goos: darwin
goarch: amd64
pkg: go.dedis.ch/dela/core/execution/benchmark
BenchmarkEVM-4                  1000000000               0.00151 ns/op
BenchmarkEVMNetwork-4           1000000000               0.0302 ns/op
PASS
```

### DIFF
--- a/core/execution/benchmark/mod_test.go
+++ b/core/execution/benchmark/mod_test.go
@@ -13,13 +13,13 @@ import (
 
 const iterations = 50
 
-func BenchmarkLocal(b *testing.B) {
-	testWithAddr(b, "127.0.0.1:12346")
-}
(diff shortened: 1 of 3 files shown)


gnarula:
I've also added a network benchmark that runs the solidity contract over tcp. The TCP server is in `evm/network` and listens on `127.0.0.1:12347` by default. Make sure to set `UNIKERNEL_TCP` environment variable correctly to run the benchmark

PR #186, opened 2021-05-07
Fixes tree search

nkcr:
There was a bug leading to tree root mismatch during consensus. This was caused by the tree search implementation that changes the tree. This PR updates the search to keep the tree state.

It also adds a new integration test and updates the display of the value contract.

### DIFF
--- a/contracts/value/mod.go
+++ b/contracts/value/mod.go
@@ -179,7 +179,7 @@ func (c valueCommand) write(snap store.Snapshot, step execution.Step) error {
 
 	c.index[string(key)] = struct{}{}
 
-	dela.Logger.Info().Str("contract", ContractName).Msgf("setting %s=%s", key, value)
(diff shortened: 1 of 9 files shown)


nkcr:
@Gilthoniel I would be very interested in having your opinion about the changes I made in `core/store/hashtree/binprefix/tree.go`, that would be awesome 🙂

For a bit of context, there was a bug during the consensus: if a node would perform a tree search, then that node could end up having a different root hash compared to the nodes that wouldn't do a search. So that's why I updated the search to not change `disknodes` to `leafnodes` during the search, which I guess was an optimization you implemented.

gnarula:
I looked this up and think it's because the follower nodes' staging tree is using a cached hash value for a `LeafNode`. `LeafNode::Insert` doesn't reset the hash on a value update.

TODO: I'm going to write a unit test to confirm this. So far 100 runs with `n.hash = nil` in the if block doesn't fail the integration test ^_^

PR #189, opened 2021-05-12
Evoting 

A-Ibm:


A-Ibm:
### **Remaining issue** 

The verification of the proof in the Neff Shuffle protocol fails sometimes, follow these instructions to reproduce the bug : 

1. run `go install` in _cli/node/memcoin_ and _cli/crypto_
2. cd to the root of the repo
3. run 3 nodes : `LLVL=info memcoin --config /tmp/node1 start --port 2001` , `LLVL=info memcoin --config /tmp/node2 start --port 2002` ,  `LLVL=info memcoin --config /tmp/node3 start --port 2003`
4. run the script `3nodes.sh`
5. run the following command : `memcoin --config /tmp/node1 e-voting scenarioTest --member $(memcoin --config /tmp/node1 dkg export) --member $(memcoin --config /tmp/node2 dkg export) --member $(memcoin --config /tmp/node3 dkg export)`

If you are lucky enough, you should see some` PROOF FAILED !!!!!!!!` printed in the CLI of nodes, if not, simply run the command at step 5 until a proof verification fails.

6. delete temp folders of _node1_ _node2_ and _node3_

A-Ibm:
### **Remaining Issue** 

I mentioned during our previous meeting we still have a nonce issue. We set the nonce of the newly pushed transaction by incrementing the nonce of the last transaction, however, at some point, the nonce is always incorrect and we end up not being able to push transactions anymore.

I thought we might have a concurrency issue when fetching the last block, but you were right, [fetching the last block](https://github.com/dedis/dela/blob/d-voting-code-refactoring/core/ordering/cosipbft/blockstore/disk.go#L242) is indeed thread safe. 

After some investigation, the last block happens to be filled with only rejected transactions. Hence, I fetch all the blocks until I find one filled with accepted transaction and I set nonce = max nonce + 1. You can find the code [here](https://github.com/dedis/dela/blob/d-voting-code-refactoring/contracts/evoting/controller/action.go#L1299). 

Everything seems to be working fine (apart from the [proof verification](https://github.com/dedis/dela/pull/189#issuecomment-840532035))

dkg/pedersen/controller/action.go
```diff
@@ -0,0 +1,450 @@
...
+func (a *initHttpServerAction) Execute(ctx node.Context) error {
+	portNumber := ctx.Flags.String("portNumber")
+
+	var actor dkg.Actor
+	err := ctx.Injector.Resolve(&actor)
+	if err != nil {
+		return xerrors.Errorf("failed to resolve actor: %v", err)
+	}
+
+	http.HandleFunc(getPublicKeyEndPoint, func(w http.ResponseWriter, r *http.Request) {
```
nkcr, line 218: should use the `mino/proxy`

dkg/pedersen/controller/action.go
```diff
@@ -0,0 +1,450 @@
...
+		_, err = fmt.Fprint(w, string(pubkeyBuf))
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+	})
+
+	plaintext := "hello"
+
+	K, C, _, err := actor.Encrypt([]byte(plaintext))
```
nkcr, line 241: don't use uppercase: 
k, c, kBuff, cBuff

dkg/pedersen/controller/action.go
```diff
@@ -0,0 +1,450 @@
...
+		}
+
+		_, err = fmt.Fprint(w, string(pubkeyBuf))
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+	})
+
+	plaintext := "hello"
```
nkcr, line 239: should be passed in the http body

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/opentracing/opentracing-go v1.2.0
 	github.com/rs/xid v1.2.1
 	github.com/rs/zerolog v1.19.0
+	github.com/satori/go.uuid v1.2.0
```
nkcr, line 13: As we're already using `"github.com/rs/xid"` for generating GUID in mino, I would stick with it and not add a new dependency.

private.key
```diff
@@ -0,0 +1 @@
+K�8�CS�}�t�^��ՙ5K�!��J���
```
nkcr, line 1: this file shouldn't be there

nkcr (changes requested):
That's a lot to review there. This is a first round and I haven't looked at the neff shuffle implementation at all. It will come with a followup review. 

Please update this branch with master, which now includes a new action that will provide you a lot of hints (that's sonarcloud).

I see you put some efforts in the tests, thank for that! Otherwise I let you go through my comments. I found the code in general clear and readable, but there is still room for a lot of refactoring, especially in `evoting/controller/action.go` 😀 .

nkcr (changes requested):
Here is the rest of the review.

shuffle/neff/handler.go
```diff
@@ -0,0 +1,363 @@
...
+					Hex("id", tx.GetID()).
+					Msg("transaction included in the block")
+
+				accepted, msg := res.GetStatus()
+
+				if !accepted {
+					notAccepted = true
+					dela.Logger.Info().Msg("Denied : " + msg)
+					break
+				} else {
```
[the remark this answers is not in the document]
ineiti: there is no golint? I think it catches those...

nkcr:
Implementation moved in https://github.com/dedis/d-voting.

PR #197, opened 2021-12-16
Allows port 0 in integration test

### LINKED ISSUE #196: It should be possible to pass a "0" port to minogrpc
Passing port "0" with the standard tcp library in go automatically uses a random free port. It should therefore be able to do that with minogrpc.

nkcr:
A misuse of the mino interface prevented the use of a port 0. In fact a node's address should be retrieved with the mino interface.

closes #196


jbsv (approved):
Thanks

PR #203, opened 2022-02-24
Changes AGPL to BSD-3 licenses

nkcr:


pierluca (changes requested):
Also need to change the README.md which mentions AGPL

pierluca (approved):

PR #206, opened 2022-03-07
Adds Prometheus metrics and handler

nkcr:
Adds Prometheus metrics and updates the proxy to run a simple Prometheus server.


core/ordering/cosipbft/pbft/mod.go
```diff
@@ -53,6 +55,29 @@ func (s State) String() string {
 	}
 }
 
+// defines prometheus metrics
+var (
+	promBlocks = promauto.NewGauge(prometheus.GaugeOpts{
+		Name: "dela_cosipbft_blocks_total",
+		Help: "total number of blocks",
+	})
```
pierluca, line 63: This looks like it should be a counter, rather than a gauge.
We only expect it to increase, right ?

core/ordering/cosipbft/pbft/mod.go
```diff
@@ -53,6 +55,29 @@ func (s State) String() string {
...
+
+	promTxs = promauto.NewGauge(prometheus.GaugeOpts{
+		Name: "dela_cosipbft_transactions_block",
+		Help: "total number of transactions in the last block",
+	})
```
pierluca, line 68: the gauge is only going to be "collected" (i.e. sampled) once in a while by prometheus. We're going to miss a lot of values between samplings. It might make more sense to have a histogram here.

It might also make sense to have a counter, counting all transactions (independently of the block), but that would fit a separate need (measuring the level of "activity" in the system)

core/ordering/cosipbft/pbft/mod.go
```diff
@@ -53,6 +55,29 @@ func (s State) String() string {
...
+
+	promRejectedTxs = promauto.NewGauge(prometheus.GaugeOpts{
+		Name: "dela_cosipbft_transactions_rejected_block",
+		Help: "total number of rejected transactions in the last block",
+	})
```
pierluca, line 73: should we only look at the last block (and possibly use a histogram), or the total number of reject transactions (of all past blocks) and use a counter for this ?

core/ordering/cosipbft/pbft/mod.go
```diff
@@ -53,6 +55,29 @@ func (s State) String() string {
...
+
+	promLeader = promauto.NewGauge(prometheus.GaugeOpts{
+		Name: "dela_cosipbft_leader",
```
pierluca, line 76: This actually looks like it should be a gauge, I won't be annoying here 😸👍

mino/proxy/http/controller/mod.go
```diff
@@ -32,6 +34,17 @@ func (m minimal) SetCommands(builder node.Builder) {
 		Value:    defaultAddr,
 	})
 	sub.SetAction(builder.MakeAction(startAction{}))
+
+	sub = cmd.SetSubCommand("prom")
```
pierluca, line 38: my understanding is that here we're only configuring the metrics path, rather than the port on which we're exposing the metrics.
We may not want to expose metrics on the same port as the application-level data.
If they're on the same port, firewalls can't (trivially) let different sources access one but not the other.
What do you think?

pierluca (commented):
Here's a few thoughts

core/ordering/cosipbft/pbft/mod.go
nkcr: True, but we can't easily initialize it if there are already some blocks. The counter doesn't let you set a value.

mino/proxy/http/controller/mod.go
nkcr: I put this command as a convenience to have a server quickly up and running. For a production case, the one where firewall rules matter for example, then it is expected that the user implement its own server.

core/ordering/cosipbft/pbft/mod.go
pierluca: It depends what we want to measure with this: total blocks or rate of adding new blocks ? 

A counter increments over the lifecycle of the application, which means we can always look at the rate (i.e. blocks/s).
But you're right that if the idea is to monitor the total, then a gauge would be better.

mino/proxy/http/controller/mod.go
pierluca: Would that then be the case in d-voting, for instance?

nkcr: Yes, in the case of d-voting we will add a small http server in our implementation to handle the Prometheus metrics. The idea is to keep Dela simple, with minimal functionalities.

Makefile
```diff
@@ -2,11 +2,12 @@ generate:
 	go get -u github.com/golang/protobuf/protoc-gen-go@v1.3.5
 	go generate ./...
 
+# Some packages are exluded from staticcheck due to deprecated warnings: #208.
 lint:
 	# Coding style static check.
 	@go get -v honnef.co/go/tools/cmd/staticcheck
 	@go mod tidy
-	staticcheck ./...
+	staticcheck `go list ./... | grep -Ev "(go\.dedis\.ch/dela/internal/testing|go\.dedis\.ch/dela/mino/minogrpc/ptypes)"`
```
pierluca, line 10: is it a good idea to hide linting warnings? It seems to me we're increasing the danger of forgetting them

core/ordering/cosipbft/pbft/mod.go
```diff
@@ -53,6 +55,41 @@ func (s State) String() string {
...
+	promRejectedTxs = promauto.NewHistogram(prometheus.HistogramOpts{
+		Name:    "dela_cosipbft_transactions_rejected_block",
+		Help:    "total number of rejected transactions in the last block",
+		Buckets: []float64{0, 1, 2, 3, 5, 8, 13, 20, 30, 50, 100},
+	})
+
+	promRejectedTxsTot = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "dela_cosipbft_transactions_rejected_tot",
+		Help: "total number of rejected transactions in history",
+	})
```
pierluca, line 85: if you're using a histogram, you don't need a counter for the same values.
Because histograms are implemented as time-series (one per bucket), you can just sum across all buckets to have the total.

core/ordering/cosipbft/pbft/mod.go
```diff
@@ -53,6 +55,41 @@ func (s State) String() string {
...
+	promTxs = promauto.NewHistogram(prometheus.HistogramOpts{
+		Name:    "dela_cosipbft_transactions_block",
+		Help:    "total number of transactions in the last block",
+		Buckets: []float64{0, 1, 2, 3, 5, 8, 13, 20, 30, 50, 100},
+	})
+
+	promTxsTot = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "dela_cosipbft_transactions_tot",
+		Help: "total number of transactions in history",
+	})
```
pierluca, line 74: I think the same histogram/counter remark applies here.

Makefile
nkcr: I created an issue to keep that in mind. In fact this is an error that fails the check, not just a warning.

pierluca (approved):

PR #207, opened 2022-03-07
Adds bash scripts and documentation to help with manual testing

jbsv:
Added high level scripts to simplify manual testing of setting up 4 nodes and adding transactions to the chain.


pierluca (approved):
I think this is "good enough" for now, but I would consider making the scripts more robust in a future step, i.e. validating that they're called with the right parameters, that they abort if a command fails, etc.

nkcr (approved):

PR #211, opened 2022-03-23
Updates server TLS minimum certificate

nkcr:
- updates grpc version


pierluca (approved):
Hoping the gRPC upgrade doesn't break anything 😅

PR #212, opened 2022-04-11
Adds public address on minogrpc

### LINKED ISSUE #162: Minogrpc certificate: hostname

nkcr:
The actual implementation won't work for nodes behind a NAT proxy. This PR adds an update for minogrpc to separate the "local address" and the "node address". The later one is used as the public address, while the former one is the address the node listens on.

close #162


mino/minogrpc/controller/mod.go
```diff
@@ -52,11 +52,19 @@ func NewController() node.Initializer {
...
-			Usage: "set the port to listen on",
-			Value: 2000,
+		cli.StringFlag{
+			Name:  "listen",
+			Usage: "set the address to listen on",
+			Value: "0.0.0.0:2000",
+		},
+		cli.StringFlag{
+			Name:     "public",
+			Usage:    "sets the public node address. By it default uses the same as listen",
```
ineiti, line 65:
```suggestion
			Usage:    "sets the public node address. By default it uses the same as --listen",
```

mino/minogrpc/mod.go
```diff
@@ -144,15 +145,23 @@ func WithRandom(r io.Reader) Option {
 }
 
 // NewMinogrpc creates and starts a new instance. it will try to listen for the
-// address and returns an error if it fails.
-func NewMinogrpc(addr net.Addr, router router.Router, opts ...Option) (*Minogrpc, error) {
-	socket, err := net.Listen(addr.Network(), addr.String())
+// address and returns an error if it fails. "listen" is the local address,
+// while "public" is the public node address. If public is empty it uses the
+// local address.
+func NewMinogrpc(listen net.Addr, public string, router router.Router, opts ...Option) (*Minogrpc, error) {
```
ineiti, line 151: What is your stability guarantees for public APIs?

mino/minogrpc/mod.go
```diff
@@ -144,15 +145,23 @@ func WithRandom(r io.Reader) Option {
...
 
+	if public == "" {
+		public = socket.Addr().String()
+	}
+
+	dela.Logger.Info().Msgf("public address is: %s", public)
+
 	tmpl := minoTemplate{
-		myAddr: session.NewAddress(socket.Addr().String()),
+		myAddr: session.NewAddress(public),
```
ineiti, line 164: Don't you want to make sure that `public` is a valid address here? It's user-input, so I'd verify it before putting it in a structure.

jbsv (approved):
yeah, maybe validate the `public` against a regex if it's different from "". Otherwise LGTM.

PR #213, opened 2022-04-27
Adds an action to delete certificates

nkcr:
Sometimes we need to remove an existing certificate from a node. This PR adds a `memcoin minogrpc certificates rm --address <addr>` CLI command to do that.
This is especially convenient when we need to regenerate a node's certificate, for example when its public address has changed.

### DIFF
--- a/mino/minogrpc/controller/mod_test.go
+++ b/mino/minogrpc/controller/mod_test.go
@@ -23,7 +23,7 @@ func TestMiniController_Build(t *testing.T) {
 	call := &fake.Call{}
 	ctrl.SetCommands(fakeBuilder{call: call})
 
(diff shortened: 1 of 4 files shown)


mino/minogrpc/controller/mod_test.go
```diff
@@ -23,7 +23,7 @@ func TestMiniController_Build(t *testing.T) {
 	call := &fake.Call{}
 	ctrl.SetCommands(fakeBuilder{call: call})
 
-	require.Equal(t, 17, call.Len())
+	require.Equal(t, 22, call.Len())
```
jbsv, line 26: Maybe name/document the constant ?

jbsv (approved):

PR #215, opened 2022-06-28
Updates PBFT to allow a leader catchup

### LINKED ISSUE #214: PBFT catch up: update the current leader
**This is a big issue that makes the system unusable if nodes restart.**

## Context

The leader rotation works in a round-robin fashion. Starting from node at index 0, each view change will assign the next node as the new leader, following a simple `(i+1) mod n` formula, where `i` is the current leader index and `n` the total number of nodes.

Every node maintains its local index of the current leader. A node will accept a view change only if the new proposed leader corresponds to *its* next expected leader. It also checks that blocks proposals come from *its* current leader.

## Problem

This leader index `i` starts at `0` when a node starts, and is not updated in case a node is behind the chain and does a catch up. This is a problem, as the leader may have changed without the catching up node knowing about that. The catching up node will then have an outdated current leader index `i = 0` and reject every block proposals.

Even worse, if the catching up node is node `0`, this node will think that it is the leader node, and send block proposals to all other nodes.

## How to reproduce

Use the scripts that are yet on the `manual_tests` branch:

```sh
cd test
./teststart.sh
# send a first transaction, everything fine
./testlist.sh 1 -1
# => kill node 0
# => use ctrl-B + arrows to navigate to the window
./testlist.sh 2 2
# Now node 1 is the leader, everything fine
# => restart node 0
./testlist.sh 2 3
# node 0 thinks he is the leader, he is sending proposals
```

### Update sync block

Update the sync block mechanism to include the current leader. This would mean that the catching up node is trusting the remote node that sends the catch up message to send the correct current leader.

Pros:

- Only touches the blockcync part

Cons:

- Partially solves the problem, as a bad node could send a wrong leader index

### Update the blocks content

Update blocks content to include the leader. This would allow the catching up node to check who is the leader in the last block, and update its current leader accordingly.

Pros:

- This is the most explicit way of solving this issue
- Adds interesting information to the block

Cons:

- Increases the size of blocks
- Having the leader in a block is not very relevant
- Feels like the simple and dirty solution

### Update the consensus mechanism

Even if a node doubts about the current leader, it should be convinced to update its current leader if a threshold of nodes came to a consensus on a block. This could be done in the `Finalize` step, when the block is committed. This would probably require to update the PBFT state machine to store the "rejected leader" in the prepare step, but then accept it if the Finalize step succeeds.

Note that this doesn't solve the problem in case nodes are performing a view change. Therefore, nodes should store view change proposals, and accept one once it has a threshold of proposals for a same node. Right now any proposals that does not concern the next expected leader is rejected.

Pros:

- Only touches the PBFT state machine
- Ensures a node always has a correct leader index
- Requires minimal changes

Cons:

- Adds a new state to the PBFT state machine

nkcr:
- updates `calculateThreshold` to handle the case where `f` is zero.

Close #214

### DIFF
--- a/core/ordering/cosipbft/pbft/mod.go
+++ b/core/ordering/cosipbft/pbft/mod.go
@@ -166,6 +166,10 @@ type round struct {
 	committed  bool
 	prevViews  map[mino.Address]View
 	views      map[mino.Address]View
+
+	// allows a node to catch up on a new leader
+	tentativeRound  types.Digest
+	tentativeLeader uint16
 }
 
 // AuthorityReader is a function to help the state machine to read the current
@@ -303,6 +307,11 @@ func (m *pbftsm) Prepare(from mino.Address, block types.Block) (types.Digest, er
 	_, index := roster.GetPublicKey(from)
 
 	if uint16(index) != m.round.leader {
+		// Allows the node to catchup on the leader. It rejects the proposal,
+		// but will accept this leader later if the block is finalized and
+		// synced to us.
+		m.round.tentativeRound = block.GetHash()
+		m.round.tentativeLeader = uint16(index)
 		return id, xerrors.Errorf("'%v' is not the leader", from)
 	}
 
@@ -377,6 +386,8 @@ func (m *pbftsm) Finalize(id types.Digest, sig crypto.Signature) error {
 	if err != nil {
 		return err
 	}
+
+	dela.Logger.Info().Msgf("finalize round with leader: %d", m.round.leader)
 
 	m.round.prevViews = nil
 	m.round.views = nil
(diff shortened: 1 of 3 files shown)


jbsv (approved):
I don't know pbft well enough I guess 🙈

core/ordering/cosipbft/pbft/mod.go
```diff
@@ -831,5 +849,9 @@ func (obs observer) NotifyCallback(event interface{}) {
 // be found with n = 3*f+1 where n is the number of participants.
 func calculateThreshold(n int) int {
 	f := (n - 1) / 3
+	if f == 0 {
+		return n
+	}
+
 	return 2 * f
```
ineiti, line 856:
```suggestion
	if n <= 3 {
		return n
	}
	f := (n - 1) / 3

	return 2 * f
```

But it's the same thing...

PR #221, opened 2022-08-03
Cleaned DKG implementation

pierluca:


dkg/pedersen/handler.go
```diff
@@ -89,82 +99,109 @@ func (h *Handler) Stream(out mino.Sender, in mino.Receiver) error {
...
-	// receive other messages in the meantime, like a Deal.
-	switch msg := msg.(type) {
-
-	case types.Start:
-		err := h.start(msg, deals, responses, from, out, in)
+	// TODO: Different state machine if already initialized or not
+	// if not initialized -> Deal/Response/Start
+	// else -> Decrypt/Reshare/Deal/Response
+	for {
+		// TODO: Extract context, set timeout
```
pierluca, line 110: I'll do this in a subsequent commit

dkg/pedersen/handler_test.go
```diff
@@ -42,6 +44,8 @@ func TestHandler_Stream(t *testing.T) {
 	require.EqualError(t, err, "expected Start message, decrypt request or Deal as first message, got: fake.Message")
 }
 
+// TODO: rewrite
```
pierluca, line 47: this too

dkg/pedersen/handler.go
```diff
@@ -89,97 +119,160 @@ func (h *Handler) Stream(out mino.Sender, in mino.Receiver) error {
...
+				from,
+			}
+
+		case types.DecryptRequest:
+			dela.Logger.Trace().Msgf("%v received decrypt request from %v\n", h.me, from)
+
+			return h.handleDecrypt(out, msg, from)
+
+		default:
+			dela.Logger.Error().Msgf(
```
nkcr, line 177: we usually don't use this kind of "multiline" function call style in our go code. Also, the `\n` is not necessary: 

```go
		dela.Logger.Error().Msgf("%v received an unsupported message %v from %v", 
			h.me, msg, from)
```

(same in other places)

nkcr (commented):
Left a couple of comments. In overall this all makes sense and clarifies a lot the code :)

ineiti:
I would add some comments to the PR itself for later review. Or create an issue with more description of what you're doing here and `Fixes #xyz` to the PR comment.

nkcr:
Since this has been pending for a while I'll be making the changes. I made a refactoring of DKG in the d-voting project which brings a simpler solution so I'll take the occasion to bring it there too.

dkg/pedersen/handler.go
```diff
@@ -89,97 +119,160 @@ func (h *Handler) Stream(out mino.Sender, in mino.Receiver) error {
...
-		return xerrors.Errorf("expected Start message, decrypt request or "+
-			"Deal as first message, got: %T", msg)
+	if h.startRes.IsStarting() {
+		dela.Logger.Warn().Msgf(
+			"%v ignored start request from %v as it is already"+
+				" starting\n", h.me, from)
+		return xerrors.Errorf("dkg is already starting")
 	}
 
+	h.startRes.Start()
```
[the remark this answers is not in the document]
nkcr: Just accidentally found out that this is called a TOCTOU error (https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use)

core/store/kv/default_test.go
```diff
@@ -11,8 +10,10 @@ import (
 	"golang.org/x/xerrors"
 )
 
+const delaTestDir = "dela-core-kv"
+
 func TestBoltDB_UpdateAndView(t *testing.T) {
-	dir, err := ioutil.TempDir(os.TempDir(), "dela-core-kv")
+	dir, err := os.MkdirTemp(os.TempDir(), delaTestDir)
```
pierluca, line 16: it would have been nice to make this kind of changes in a separate PR 😅

mino/minogrpc/mod.go
```diff
@@ -95,7 +95,7 @@ type Endpoint struct {
 	// that the stream session must be created. Using a sync.Map would require
 	// to use the "LoadOrStore" function, which would make us create the session
 	// each time, but only saving it the first time.
-	debugsync.RWMutex
+	sync.RWMutex
```
pierluca, line 98: why reverting away from `debugsync` ?

mino/minogrpc/mod_test.go
```diff
@@ -169,7 +169,7 @@ func TestMinogrpc_Token(t *testing.T) {
 func TestMinogrpc_GracefulClose(t *testing.T) {
 	m := &Minogrpc{
 		overlay: &overlay{
-			closer:  new(debugsync.WaitGroup),
+			closer:  new(sync.WaitGroup),
```
pierluca, line 172: same question, here and in the other files.

go.mod
```diff
@@ -13,15 +11,41 @@ require (
 	github.com/rs/zerolog v1.19.0
 	github.com/stretchr/testify v1.7.0
 	github.com/uber/jaeger-client-go v2.25.0+incompatible
-	github.com/uber/jaeger-lib v2.4.0+incompatible // indirect
```
pierluca, line 16: how can this still work ? we haven't changed our usage of jaeger, have we? 🤔

dkg/pedersen/mod_test.go
```diff
@@ -150,7 +159,7 @@ func TestPedersen_Scenario(t *testing.T) {
 	for i := 0; i < n; i++ {
 		addr := minogrpc.ParseAddress("127.0.0.1", 0)
 
-		m, err := minogrpc.NewMinogrpc(addr, nil, flat.NewRouter(minogrpc.NewAddressFactory()))
+		m, err := minogrpc.NewMinogrpc(addr, nil, tree.NewRouter(minogrpc.NewAddressFactory()))
```
pierluca, line 162: it could be nice (for now) , to keep the two scenari, one with tree routing and one with flat routing.
It shouldn't change a thing and it's the wrong place to test routing, I agree.
On the other hand, this test case actually produced interesting results, so pragmatically it might make sense to keep both (and put a comment saying that some day we should probably get rid of one).

dkg/pedersen/handler.go
```diff
@@ -119,64 +110,83 @@ func (h *Handler) Stream(out mino.Sender, in mino.Receiver) error {
 	// messages to the other nodes, and then we might get their messages before
 	// the start message.
 
-	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
+	// We make sure not additional request is accepted if a setup is in
+	// progress.
+	h.Lock()
+	if !h.startRes.Done() && h.running {
```
pierluca, line 116:
```suggestion
	if h.running {
```
wouldn't this be sufficient / cleaner ?
and if not, documenting why might be useful to future programmers :-)

dkg/pedersen/handler.go
```diff
@@ -185,144 +195,112 @@ func (h *Handler) Stream(out mino.Sender, in mino.Receiver) error {
...
 }
 
 // start is called when the node has received its start message. Note that we
 // might have already received some deals from other nodes in the meantime. The
 // function handles the DKG creation protocol.
-func (h *Handler) start(ctx context.Context, start types.Start,
-	from mino.Address, out mino.Sender) error {
-
-	dela.Logger.Info().Msgf("%v is starting a DKG", h.me)
+func (h *Handler) start(start types.Start, deals, resps *list.List, from mino.Address,
```
pierluca, line 225: question for my understanding: how are we going to stop this when the stream gets closed, if they don't share a context ?

dkg/pedersen/handler.go
```diff
@@ -185,144 +195,112 @@ func (h *Handler) Stream(out mino.Sender, in mino.Receiver) error {
...
-				"pubKey: %d != %d", len(start.GetAddresses()),
-			len(start.GetPublicKeys()),
-		)
+	if len(start.GetAddresses()) != len(start.GetPublicKeys()) {
+		return xerrors.Errorf("there should be as many participants as "+
+			"pubKey: %d != %d", len(start.GetAddresses()), len(start.GetPublicKeys()))
 	}
-	n := len(participants)
-	ctx, cancel := context.WithTimeout(ctx, time.Duration(n*n)*time.Second)
-	defer cancel()
```
pierluca: shouldn't we time out after an arbitrary long amount of time, if the DKG is completely stalled ? 
It seems brittle not to have a max timeout for the DKG, to at least return "this failed"

dkg/pedersen/handler.go
```diff
@@ -335,81 +313,86 @@ func (h *Handler) sendDeals(ctx context.Context, out mino.Sender,
...
-			dela.Logger.Error().Msgf("%s timed out while sending deals", h.me)
-			return xerrors.Errorf("timed out while sending deals")
+	for numReceivedDeals < len(h.startRes.participants)-1 {
+		h.Lock()
+		deal := deals.Front()
+		if deal != nil {
+			deals.Remove(deal)
 		}
-	}
+		h.Unlock()
```
pierluca, line 341: I'm still very uncomfortable with this whole setup.
I'd at least consider creating a `syncQueue` structure that wraps around this logic, so that the DKG doesn't have to handle the lock itself, but it gets hidden behind a `.Pop()` call of some kind.

dkg/pedersen/handler.go
```diff
@@ -335,81 +313,86 @@ func (h *Handler) sendDeals(ctx context.Context, out mino.Sender,
...
 
-	dela.Logger.Debug().Msgf("%s sent all its deals", h.me)
+		if deal == nil {
+			time.Sleep(retryTimeout)
+			continue
```
pierluca, line 345: This is where a channel abstraction would make me feel more comfortable 😅

dkg/pedersen/handler.go
```diff
@@ -335,81 +313,86 @@ func (h *Handler) sendDeals(ctx context.Context, out mino.Sender,
...
+		}
+		h.Unlock()
 
-		case <-ctx.Done():
-			dela.Logger.Error().Msgf("%s timed out while receiving deals", h.me)
-			return xerrors.Errorf("timed out while receiving deals")
+		if resp == nil {
+			time.Sleep(retryTimeout)
+			continue
 		}
```
pierluca, line 375: ditto

dkg/pedersen/handler.go
```diff
@@ -185,144 +195,112 @@ func (h *Handler) Stream(out mino.Sender, in mino.Receiver) error {
...
+	// orchestrator, so that it can process decrypt requests right away.
+	h.startRes.SetDistKey(distKey.Public())
+
+	h.Lock()
+	h.privShare = distKey.PriShare()
+	h.running = false
+	h.Unlock()
+
+	done := types.NewStartDone(distKey.Public())
+	err = <-out.Send(done, from)
```
pierluca, line 285: it seemed sort of natural (and easy) to put all of this in `announceDkgPublicKey` , given the different levels of abstraction.
What do you think ?

core/store/kv/default_test.go
nkcr: Yes, I would've definitely left the update to Go 1.19 to another PR 🙃.

Makefile
```diff
@@ -10,7 +10,7 @@ lint:
 	staticcheck `go list ./... | grep -Ev "(go\.dedis\.ch/dela/internal/testing|go\.dedis\.ch/dela/mino/minogrpc/ptypes)"`
 
 vet:
-	@echo "⚠️ Warning: the following only works with go >= 1.14" && \
+	@echo "⚠️ Warning: the following only works with go >= 1.19" && \
```
nkcr, line 13: It shouldn't be changed. This is related to something that `mcheck` uses that requires go >= 1.14.

dkg/pedersen/handler.go
nkcr: yes, it works too

nkcr:
@pierluca I added back buffered channels and a mean to cancel the start.
We should eventually move the `cryChan` datastructure in another package but I wouldn't add any more delays to merge this PR. So I would advise we do that in a subsequent PR.

dkg/pedersen/handler.go
```diff
@@ -26,13 +26,57 @@ import (
...
+}
+
+func (c *cryChan[T]) pop(ctx context.Context) (t T, err error) {
+	select {
+	case el := <-c.c:
+		return el, nil
+	case <-ctx.Done():
+		return t, ctx.Err()
+	}
+}
```
pierluca, line 74: when we move this out, I'd have a few comments regarding API design decisions that I'd change. For this PR, this is good enough.

Writing down my (as-of-yet-incoherent) thoughts so I don't forget later on: 
- I'd consider exposing the channel directly for reading, as a `context` might not always be flexible enough
- I wonder if we should set a global `write timeout` to the channel (much like `SetWriteDeadline`), or be able to provide individual timeouts for writes
- The "crying" should probably provide a unique identifier (say, a go routine ID) so that we can make sense of logs and distinguish between multiple unique blocked channels.

dkg/pedersen/handler.go
```diff
@@ -26,13 +26,57 @@ import (
...
+type cryChan[T any] struct {
+	c   chan T
+	log zerolog.Logger
+}
+
+func (c *cryChan[T]) push(e T) {
+	start := time.Now()
+	select {
+	case c.c <- e:
+	case <-time.After(time.Millisecond * 100):
```
pierluca, line 59: Since this is just a fail-safe, and 100ms is in the order of magnitude of a minor computer freeze, I'd consider increasing this to the range of seconds / tens of seconds.

dkg/pedersen/handler.go
```diff
@@ -123,8 +167,11 @@ func (h *Handler) Stream(out mino.Sender, in mino.Receiver) error {
 	}
 	h.Unlock()
 
-	deals := list.New()
-	responses := list.New()
+	deals := newCryChan[types.Deal](chanSize)
+	responses := newCryChan[*pedersen.Response](chanSize)
```
pierluca, line 171: is there a reason in one case we use `*pedersen.Response` and in the other `types.Deal` ?
i.e. in one case we process the message before adding it to the channel, in the other we add the raw network message to the channel ? 
i think that's a regression, we'd made this uniform :)

dkg/pedersen/handler.go
```diff
@@ -123,8 +167,11 @@ func (h *Handler) Stream(out mino.Sender, in mino.Receiver) error {
...
+
+	globalCtx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
pierluca, line 174: we'll know about it, but this will still block indefinitely right?
there's no timeout mechanism in place, shouldn't we keep one?

especially when we run this with lots of nodes, it's hard to tell if we're stuck or still doing actual useful work.

pierluca (commented):
A couple of things, but I'm a lot more comfortable with this  already !

dkg/pedersen/handler.go
pierluca: @nkcr :innocent:

dkg/pedersen/handler.go
nkcr: It blocks until the function exits. This context's job is to make sure we don't have a stale go routine in case the function exits badly. Having a timeout on the receiving loop is another question. It should be the orchestrator's responsibility (the one that starts the setup) to terminate the stream after some timeout, thus making the function exit.

dkg/pedersen/handler.go
nkcr: I'm happy with both 🤷‍♂️

pierluca: I think there was a misunderstanding 🤦‍♂️ 

I have no issue with the pointer, but rather, where we receive the messages:
- in the response case, we convert them immediately (in `func Stream`, case `types.Response`) to a Kyber type ( `pedersen.Respones` )
- in the deal case, we store the `types.Deal` in `func Stream`, but then convert it to a Kyber type (`pedersen.Deal`) in `func handleDeal`.

Basically, it would make sense for the internal API to either be consistent about using `pedersen` types or `types` types 😄

pierluca (commented):
LGTM !

PR #222, opened 2022-08-16
Resharing

Mahsa-Bastankhah:
I added the reshaping functionality to dkg

### DIFF
--- a/dkg/mod.go
+++ b/dkg/mod.go
@@ -2,6 +2,7 @@ package dkg
 
 import (
 	"go.dedis.ch/dela/crypto"
+	"go.dedis.ch/dela/mino"
 	"go.dedis.ch/kyber/v3"
 )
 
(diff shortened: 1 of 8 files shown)


dkg/mod.go
```diff
@@ -26,5 +27,5 @@ type Actor interface {
 	Encrypt(message []byte) (K, C kyber.Point, remainder []byte, err error)
 	Decrypt(K, C kyber.Point) ([]byte, error)
 
-	Reshare() error
+	Reshare(T_new int, T_old int, addrs_new []mino.Address, pubkeys_new []kyber.Point, pubkeys_old []kyber.Point) error
```
nkcr, line 30: - is it necessary to specify the old threshold ? We should already know it
- same for the old public keys, we should already know them
- [the convention in Go is to use MixedCaps or mixedCaps rather than underscores](https://go.dev/doc/effective_go#mixed-caps)

The function should pretty much look the same as the `Setup` one:

```go
Reshare(co crypto.CollectiveAuthority, threshold int)
```

nkcr (commented):
Thank you for this nice addition :) 
Before jumping into a more in-depth review I'm leaving a first comment about the interface and waiting for your feedback.

Mahsa-Bastankhah:
Moving to a separate branch.

PR #223, opened 2022-08-17
test multi nodes messaging

jbsv:
setup a simple test as an attempt to reproduce the multi-messaging issue in TestPedersen_Scenario (with flat routing)

### DIFF
--- a/mino/minogrpc/example_test.go
+++ b/mino/minogrpc/example_test.go
@@ -3,6 +3,9 @@ package minogrpc
 import (
 	"context"
 	"fmt"
+	"github.com/stretchr/testify/require"
+	"go.dedis.ch/dela/mino/router/flat"
+	"testing"
 	"time"
 
 	"go.dedis.ch/dela/internal/tracing"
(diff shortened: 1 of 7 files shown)


mino/minogrpc/example_test.go
```diff
@@ -11,6 +14,110 @@ import (
...
+
+	for _, n := range nodes {
+		r := mino.MustCreateRPC(n, "test", exampleHandler{}, exampleFactory{})
+		rpcs = append(rpcs, r)
+	}
+
+	return rpcs
+}
+
+func createNodes(nbNodes int, useTree bool) []*Minogrpc {
```
nkcr, line 99: instead of passing a bool, you could pass the routing object.

mino/minogrpc/example_test.go
```diff
@@ -11,6 +14,110 @@ import (
...
+
+				require.True(t, from.Equal(addr))
+				require.Equal(t, msgReceived, msgSent)
+			}
+		}(i, r)
+	}
+}
+
+func generateMessages(nbNodes int) []string {
+	var messages []string
```
nkcr, line 58: A good habit is to initialize the slice at the right size if you know it in advance. This is a simple memory optimization:

```go
messages := make([]string, nbNodes)

for i := 0; i < nbNodes; i++ {
  messages[i] = fmt.Sprintf("MSG%d", i)
}

return messages
```

same bellow

mino/minogrpc/example_test.go
```diff
@@ -11,6 +14,110 @@ import (
...
+	rpcs := createRpcs(nodes)
+	exchangeCertificates(nodes)
+	players := createPlayers(nodes)
+	msgs := generateMessages(NB_NODES)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	for i, r := range rpcs {
+		go func(i int, rpc mino.RPC) {
```
nkcr, line 30: your test will exit without waiting for the messages to be sent/received. You need to wait on your goroutines termination:

```go
wg := sync.WaitGroup{}
wg.Add(NB_NODES)


	for i, r := range rpcs {
		go func(i int, rpc mino.RPC) {
			defer wg.Done()
...

wg.Wait()
```

mino/minogrpc/example_test.go
```diff
@@ -11,6 +14,110 @@ import (
 	"go.dedis.ch/dela/serde"
 )
 
+func TestRpcStreamTree(t *testing.T) {
+	NB_NODES := 200
```
nkcr, line 18: Go convention is to [not use underscores](https://go.dev/doc/effective_go#mixed-caps). Should be rather `NumNodes` or something like that.

mino/minogrpc/example_test.go
```diff
@@ -11,6 +14,110 @@ import (
 	"go.dedis.ch/dela/serde"
 )
 
+func TestRpcStreamTree(t *testing.T) {
```
nkcr, line 17: If this is a test and not an example then it will need to be moved to a _test.go file. Could be a new `scenario_test.go` file, or maybe in `mod_test.go`.

mino/minogrpc/example_test.go
```diff
@@ -11,6 +14,110 @@ import (
...
+			if err != nil {
+				panic("stream failed: " + err.Error())
+			}
+
+			for _, n := range nodes {
+				addr := n.GetAddress()
+
+				msgSent := msgs[i]
+				err = <-sender.Send(exampleMessage{value: msgSent}, addr)
+				if err != nil {
```
nkcr, line 41: you can use `require.NoError(t, err)`

PR #229, opened 2022-09-27
Remove flat routing

jbsv:
The original flat routing isn't used nor useable, so better take it off the code. To replace it, an option function has been added to tree routing constructor to be able to specify its height. Specifying a max tree height of 1 practically results in a flat routing.


mino/router/tree/mod.go
```diff
@@ -32,8 +31,17 @@ type Router struct {
 	hsFac     router.HandshakeFactory
 }
 
+// RouterOption is the signature of the option constructor
+type RouterOption func(*Router)
+
+func WithHeight(maxHeight int) RouterOption {
```
jbsv, line 37: @nkcr Is it possible to add a variable to an option function ?

nkcr: yes, that's the point :)

jbsv: Now that's magic to me!

pierluca: ditto.

PR #230, opened 2022-09-29
Adds dkg resharing

nkcr:
Mainly adapts @Mahsa-Bastankhah 's code.


dkg/pedersen/handler.go
```diff
@@ -513,5 +656,321 @@ func (h *Handler) handleDeal(ctx context.Context, msg types.Deal, out mino.Sende
...
 	return nil
 }
+
+func (h *Handler) announceDkgPublicKey(isCommonNode bool, out mino.Sender, from mino.Address) error {
+	// 6. Send back the public DKG key
+
+	var publicKey kyber.Point
+
+	isOldNode := h.startRes.GetDistKey() != nil
+	isNewNode := !isOldNode
```
pierluca, line 670:
```suggestion
	isNewNode := h.startRes.GetDistKey() == nil
```
old node doesn't seem used in this function.

dkg/pedersen/handler.go
```diff
@@ -513,5 +656,321 @@ func (h *Handler) handleDeal(ctx context.Context, msg types.Deal, out mino.Sende
...
+
+	h.log.Info().Msgf("resharing with %v", resharingRequest.GetAddrsNew())
+
+	addrsNew := resharingRequest.GetAddrsNew()
+
+	isOldNode := h.startRes.GetDistKey() != nil
+	isNewNode := !isOldNode
+
+	// By default the node is not common. Later we check
+	isCommonNode := false
```
pierluca, line 739: in this and the other functions that use these variables, wouldn't it make sense to replace this with an enum ? e.g. `nodeDkgGroup` = `old`, `new`, `common` and documenting those constants ?

dkg/pedersen/handler.go
```diff
@@ -513,5 +656,321 @@ func (h *Handler) handleDeal(ctx context.Context, msg types.Deal, out mino.Sende
 		}
 	}
 
+	dela.Logger.Debug().Msgf("%s is certified", h.me)
+
 	return nil
 }
+
+func (h *Handler) announceDkgPublicKey(isCommonNode bool, out mino.Sender, from mino.Address) error {
```
pierluca, line 664: it's a bit weird from an API perspective that `isCommonNode` gets passed in, but `isNewNode` gets computed internally.

dkg/pedersen/handler.go
```diff
@@ -513,5 +656,321 @@ func (h *Handler) handleDeal(ctx context.Context, msg types.Deal, out mino.Sende
...
+	isNewNode := !isOldNode
+
+	// if the node is new or a common node it should update its state
+	if isNewNode || isCommonNode {
+		distrKey, err := h.dkg.DistKeyShare()
+		if err != nil {
+			return xerrors.Errorf("failed to get distr key: %v", err)
+		}
+		publicKey = distrKey.Public()
+		// 7. Update the state before sending to acknowledgement to the
```
pierluca, line 679:
```suggestion
		publicKey = distrKey.Public()
		
		// 7. Update the state before sending to acknowledgement to the
```

Also, does this `7.` numbering still make sense?

dkg/pedersen/handler.go
```diff
@@ -513,5 +656,321 @@ func (h *Handler) handleDeal(ctx context.Context, msg types.Deal, out mino.Sende
...
+	// Note that a node can be old and common
+	if isOldNode {
+		expectedResponses = (resharingRequest.TNew) * len(h.startRes.GetParticipants())
+	}
+	if isCommonNode {
+		expectedResponses = (resharingRequest.TNew - 1) * len(addrsOld)
+	}
+	if isNewNode {
+		expectedResponses = (resharingRequest.TNew - 1) * resharingRequest.TOld
+	}
```
pierluca, line 806: conveniently this would be a switch if we had an enum

dkg/pedersen/handler.go
```diff
@@ -513,5 +656,321 @@ func (h *Handler) handleDeal(ctx context.Context, msg types.Deal, out mino.Sende
...
+
+	return nil
+}
+
+///////// auxiliary functions
+
+// isInSlice gets an address and a slice of addresses and returns true if that
+// address is in the slice. This function is called for cheking whether an old
+// committee member is in the new commmitte as well or not
+func isInSlice(addr mino.Address, addrs []mino.Address) bool {
```
pierluca, line 946: If we move this to a separate package, I'd adopt the more typical API `func Contains[E comparable](s []E, v E) bool`.
As a matter of fact, that's what might get integrated in the standard library at some point
https://pkg.go.dev/golang.org/x/exp/slices#Contains

pierluca (commented):
I did not have the energy to go through the whole resharing test, but I've plenty of comments on the rest to improve the code cleanliness. Nothing blocking though.

As discussed on Slack, I'd be okay to merge this nearly as-is in the interest of unblocking @Mahsa-Bastankhah ASAP, provided we come back to it soon to do the necessary clean-ups.

Let me know how you want to proceed

dkg/pedersen/handler.go
```diff
@@ -84,13 +84,92 @@ func (c *cryChan[T]) Len() int {
...
+		}
+	case certified:
+		if current != sharing && current != resharing {
+			return xerrors.Errorf("certified state must switch from sharing or resharing: %s", current)
+		}
+	case resharing:
+		if current != initial && current != certified {
+			return xerrors.Errorf("resharing state must switch from initial or certified: %s", current)
+		}
+	}
```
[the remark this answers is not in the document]
nkcr: there is no unknown case, you are refering to the string function that returns UNKNOWN in case it doesn't find the state, which should never happen.

dkg/pedersen/handler.go
```diff
@@ -209,20 +304,56 @@ func (h *Handler) Stream(out mino.Sender, in mino.Receiver) error {
...
+		case types.Reshare:
+			err = h.startRes.checkState(initial, certified, resharing)
+			if err != nil {
+				return xerrors.Errorf("bad state: %v", err)
+			}
+
+			reshares.push(msg)
+
 		case types.Response:
+			err = h.startRes.checkState(initial, sharing, certified, resharing)
```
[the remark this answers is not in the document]
nkcr: True, this is more a matter of documenting the expected behaviour.

dkg/pedersen/handler.go
```diff
@@ -513,5 +656,321 @@ func (h *Handler) handleDeal(ctx context.Context, msg types.Deal, out mino.Sende
 		}
 	}
 
+	dela.Logger.Debug().Msgf("%s is certified", h.me)
```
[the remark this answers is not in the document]
nkcr: true

dkg/pedersen/mod.go
```diff
@@ -268,7 +272,111 @@ func (a *Actor) Decrypt(K, C kyber.Point) ([]byte, error) {
...
+		if i != 0 && !dkgPubKeys[i-1].Equal(doneMsg.GetPublicKey()) {
+			return xerrors.Errorf("the public keys does not match: %v", dkgPubKeys)
+		}
+	}
 	return nil
 }
+
+// Gets the list of the old committee members and new committee members and
+// returns the new committee members that are not common
+func subtractOfTwoSlices(addrsSlice1 []mino.Address, addrsSlice2 []mino.Address) []mino.Address {
```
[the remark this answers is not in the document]
nkcr: No, and for the time being I wouldn't do that, especially given the probability that the standard golang library will eventually provide it.

pierluca (approved):
As discussed, merging as-is and we'll make a second pass on this.

dkg/pedersen/handler.go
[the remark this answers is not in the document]
pierluca: sorry, that was a "3 am"-quality review 😅

dkg/pedersen/handler.go
```diff
@@ -656,34 +656,29 @@ func (h *Handler) handleDeal(ctx context.Context, msg types.Deal, out mino.Sende
...
 
-	var publicKey kyber.Point
-
-	isOldNode := h.startRes.GetDistKey() != nil
-	isNewNode := !isOldNode
+	publicKey := h.startRes.GetDistKey()
 
 	// if the node is new or a common node it should update its state
-	if isNewNode || isCommonNode {
+	if publicKey == nil || isCommonNode {
```
pierluca, line 668: This is waaaay less explicit than `isNewNode` though :-/

mino/minoch/rpc.go
```diff
@@ -207,9 +211,19 @@ func (c RPC) Stream(ctx context.Context, memship mino.Players) (mino.Sender, min
...
+						default:
+							fmt.Println("FULL!! (orch)")
+							orchRecv.out <- env
+						}
 					} else {
-						outs[to.String()].out <- env
+						select {
+						case outs[to.String()].out <- env:
+						default:
+							fmt.Println("FULL!!")
```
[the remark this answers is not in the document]
jbsv: it's not possible to use cryChan here, as it would require to change the struct of a receiver, which then breaks the receive function that selects on different channels...

PR #235, opened 2022-11-22
Dkg v2 improved

jbsv:


dkg/pedersen/handler.go
```diff
@@ -28,13 +28,18 @@ import (
...
-// const chanSize = 10000
-
 // constant used in the logs
 const newState = "new state"
+const badState = "bad state: %v"
+const failedState = "failed to switch state: %v"
+
+// enumeration of the node type
+const (
+	unknownNode = iota
```
nkcr, line 38: you can only declare `iota` once, expressions are implicitly repeated in a paren­thesized const declaration

dkg/pedersen/handler.go
```diff
@@ -670,12 +675,12 @@ func (h *Handler) handleDeal(ctx context.Context, msg types.Deal,
 	return nil
 }
 
-func (h *Handler) finalizeReshare(ctx context.Context, isCommonNode bool, out mino.Sender, from mino.Address) error {
+func (h *Handler) finalizeReshare(ctx context.Context, nodeType int, out mino.Sender, from mino.Address) error {
```
nkcr, line 678: would be even better to declare a type:

```
type nodeType byte
const (
    unknownNode nodeType = iota
    oldNode
    commonNode
...
)
...
func (h *Handler) finalizeReshare(ctx context.Context, t nodeType, out mino.Sender, from mino.Address) error {
```

nkcr (commented):
Looks good, just a minor comment

dkg/pedersen/handler.go
jbsv: Didn't know, but agree. What a cool language. thanks

Issue #238, opened 2022-12-12
Performance is limited by verification of already-verified transactions


pierluca:
As shown in [profile.pdf](https://github.com/dedis/dela/files/10206896/profile001.pdf), a lot of the time is spent in the bn256 verification function. There's a legitimate need for verification, but we also noticed that this can happen due to deserialization of already-verified blocks and this needlessly slows down the blockchain.

pierluca:
### Meeting notes , 12/12/2022

#### Participants:
@jbsv , @nkcr , @pierluca 

#### Issues identified

- In the ordering `cosipbft` implementation, all links are verified from the genesis block until the latest block. It should be enough to verify the latest blocks and their links.
- In the service `GetProof` call, all the blocks are deserialized whereas only the last block (and all past links) is really needed.

#### Outcomes

- The above optimizations have been identified (and implemented on the fly) as part of the meeting, we expect performance improvements in the range of 80-90% for typical scenarios.
- @nkcr is fixing the tests and we'll reconvene if the performance hasn't been massively improved.
- we'll subsequently implement some benchmarks to ensure non-regression of the performance.

PR #239, opened 2022-12-13
Optimizes some verification operations

nkcr:
- The chain verification verifies only from a given Digest, in case we trust blocks already stored
- GetProof will get links up to the last one, instead of blocklinks, sparing blocks unmarshalling and verification

Fix https://github.com/dedis/dela/issues/238

### DIFF
--- a/core/ordering/cosipbft/blockstore/disk_test.go
+++ b/core/ordering/cosipbft/blockstore/disk_test.go
@@ -145,6 +145,20 @@ func TestInDisk_GetChain(t *testing.T) {
 
 	store.fac = badLinkFac{}
 	_, err = store.GetChain()
+	require.EqualError(t, err, fake.Err("while reading database: while scanning: link malformed"))
+}
+
+func TestInDisk_GetChain_BadBlockLink(t *testing.T) {
+	db, clean := makeDB(t)
+	defer clean()
+
+	store := NewDiskStore(db, makeBlockFac())
+
+	err := store.Store(makeLink(t, types.Digest{}, types.WithIndex(0)))
+	require.NoError(t, err)
+
+	store.fac = badLinkFac{}
+	_, err = store.GetChain()
 	require.EqualError(t, err, fake.Err("while reading database: while scanning: block malformed"))
 }
 
--- a/core/ordering/cosipbft/blockstore/disk_test.go
+++ b/core/ordering/cosipbft/blockstore/disk_test.go
@@ -264,6 +264,10 @@ func (badLinkFac) BlockLinkOf(serde.Context, []byte) (types.BlockLink, error) {
 	return nil, fake.GetError()
 }
 
+func (badLinkFac) LinkOf(serde.Context, []byte) (types.Link, error) {
+	return nil, fake.GetError()
+}
+
 type badLink struct {
 	types.BlockLink
 }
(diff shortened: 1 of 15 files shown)


core/ordering/cosipbft/proof.go
```diff
@@ -40,9 +40,9 @@ func (p Proof) GetValue() []byte {
 }
 
 // Verify takes the genesis block and the verifier factory to verify the chain
-// up to the latest block.
+// up to the latest block. It verifies the whole chain.
```
pierluca, line 43: Indeed, and as such this function doesn't scale.
Is it a utility function that's never in the hot path or is this going to be our next bottleneck? 😅

pierluca (commented):
LGTM overalls, just a minor question.

core/ordering/cosipbft/proof.go
nkcr: As long as we don't have skipchains with higher links, yes, this is going to be a bottleneck. But it should be significantly faster with the link/bocklink optimization.

pierluca: Right, I wasn't sure this was the same `GetProof` we were talking about yesterday. I must be getting old 😅 
Fair enough, we'll look at higher links Reasonably Soon (TM).

pierluca (approved):
Approved on my side, but I'll let @jbsv review this too.

nkcr:
Added a benchmark. Turns out this is now 1.8x times faster 🚀 
[Dela optimization benchmark - Sheet3.pdf](https://github.com/dedis/dela/files/10217254/Dela.optimization.benchmark.-.Sheet3.pdf)

core/ordering/cosipbft/types/chain_test.go
```diff
@@ -203,48 +203,108 @@ func TestChain_Verify(t *testing.T) {
 	genesis, err := NewGenesis(ro)
 	require.NoError(t, err)
 
-	c := NewChain(makeLink(t, genesis.digest), nil)
+	c := NewChain(makeLink(t, genesis.digest, Digest{}), nil)
 
-	err = c.Verify(genesis, fake.VerifierFactory{})
+	err = c.Verify(genesis, genesis.GetHash(), fake.VerifierFactory{})
```
jbsv, line 208: Couldn't we call 'genesis.GetHash() only once and save it in a variable to speed up testing ?

nkcr:
And here is the profiling report:
[profile001.pdf](https://github.com/dedis/dela/files/10218148/profile001.pdf)

core/ordering/cosipbft/blockstore/disk_test.go
```diff
@@ -143,6 +143,20 @@ func TestInDisk_GetChain(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, uint64(1), chain.GetBlock().GetIndex())
 
+	store.fac = badLinkFac{}
+	_, err = store.GetChain()
```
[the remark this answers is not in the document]
nkcr: Yes. This is still checking for a link, not a block link as bellow.

core/ordering/cosipbft/types/chain_test.go
nkcr: `GetHash()` is an attribute, we won't gain much by saving it in a variable.

Issue #236, opened 2022-11-24
Documentation is not readily navigable in an obvious way


bford:
Problem 1: The README.md at the top of the tree links to [GitHub pages](https://dedis.github.io/dela/#/), but that page has just a bare "what is Dela" landing page with no obvious further links to the actual documentation.  It looks like this page needs to incorporate or link to 'docs/sidebar.md' or some equivalent, but doesn't yet.

Problem 2: A lot of the "real" detailed API documentation is GoDoc, but there's no obvious link leading to that.  It looks to me like the top-level README.md, as well as docs/sidebar.md, should probably offer links to [the godocs](https://godocs.io/go.dedis.ch/dela) for detailed API documentation.

bford:
To update this comment after realizing that the documentation page does actually have documentation but you have to scroll down to get to it...  The landing page is pretty but a bit *too* minimalistic, with no actual visual hint about where one can go from it and how.  There are at least two obvious solutions:

- Make the sidebar present throughout the *whole* vertical page, not just below the landing part.  If you want the landing page to have the "bare" look to start with, the sidebar could be closed initially so that only the little navigation button appears in the upper left corner that pops out the sidebar.
- Make the Dela name, logo, and most/all of the text underneath it into a link or button that simply scrolls down to the start of the "real content" below the landing page.  Just make sure that clicking just about anywhere does something good, namely getting the user to the real content.

Further, the "Go lint" etc bar could probably be moved to the bottom, or even be moved off the landing page to the first main content page below - I don't see why a reader needs to see Dela's "Go lint" status before they can see anything else about what it is.

PR #243, opened 2023-01-30
Remove coverpage to allow long sidebar

jbsv:
Related to https://github.com/dedis/dela/issues/236, here is an attempt to improve browsing the DELA doc page.


pierluca:
@jbsv how can I see the output ? 😅

jbsv:
from your local git copy of dela: `docsify serve docs`, which implies that you install `docsify-cli` on your machine

dkg/pedersen/dkg_test.go
```diff
@@ -1064,8 +1064,6 @@ func getCertified(t *testing.T) (*pedersen.DistKeyGenerator, *pedersen.Response)
 }
 
 type fakeDKGService struct {
-	dkgService
```
pierluca, line 1067: are you sure about this in this PR ? 😅

pierluca:
Okay, I've been able to run it. Thanks for the instructions.

Couple of points of feedback:
1. "General Purpose", "Code" and "Instructions" in the menu seem to be not clickable ? It's a bit unclear.
If they're supposed to be headers, some vertical spacing might be helpful, but I'm not sure they're very helpful. 
If we just rename "Guideline" into "Code Guideline", the rest is self-explanatory.

2. The home page would stand to benefit from having the introduction text featured more prominently and the image way smaller, as right now (on my screen, at least) the home page provides 0 useful content and just one BIIIIIIIIG picture 😅 

3. The code guidelines might need (slight) updating if you want to drop the mod.go guideline.

dkg/pedersen/dkg_test.go
jbsv: yes

pierluca (approved):

PR #252, opened 2023-04-06
Fix panic in TestMiniController_FailMarshalKey_OnStart

### LINKED ISSUE #249: TestMiniController_FailMarshalKey_OnStart segfaults
https://github.com/dedis/dela/pull/246#issuecomment-1479948341

pierluca:
This is a candidate fix as it's hard to reproduce locally on my Mac, I'll be using the CI to verify.

Fixes #249

### DIFF
--- a/internal/testing/fake/store.go
+++ b/internal/testing/fake/store.go
@@ -130,7 +130,13 @@ func (tx dbTx) GetBucket(name []byte) kv.Bucket {
 
 // GetBucketOrCreate implements kv.WritableTx.
 func (tx dbTx) GetBucketOrCreate(name []byte) (kv.Bucket, error) {
-	return tx.GetBucket(name), tx.err
+	bucket, found := tx.buckets[string(name)]
+	if !found {
+		bucket = NewBucket()
+		tx.buckets[string(name)] = bucket
+	}
+
+	return bucket, tx.err
 }
 
 // GetBucketOrCreate implements store.Transaction.
--- a/mino/minogrpc/controller/mod_test.go
+++ b/mino/minogrpc/controller/mod_test.go
@@ -436,15 +436,15 @@ func (b badCurve) IsOnCurve(x, y *big.Int) bool {
 }
 
 func (b badCurve) Add(x1, y1, x2, y2 *big.Int) (x, y *big.Int) {
-	panic("unsupported")
+	panic("not implemented - should not be needed by tests")
 }
 
 func (b badCurve) Double(x1, y1 *big.Int) (x, y *big.Int) {
-	panic("unsupported")
+	panic("not implemented - should not be needed by tests")
 }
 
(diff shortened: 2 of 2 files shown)


pierluca:
Observations: The fake store implementation didn't respect the API's intentions.
It's unclear if this is the cause of the bug we observe.

pierluca:
Observation: fixing the fake store prevents the panic.

However, the test still fails and it appears the panic was caused by the test going "too far" before failing.

I'm now testing the hypothesis that the failure is due to a file existing (unexpectedly) in the filesystem.

pierluca:
Conclusion: the test failure was due to changes in the Golang crypto library.

The elliptic curve we assumed would be unsupported is now fully supported.

As such, I've modified the `badCurve` implementation to be a genuinely broken, unsupported implementation.

lourkeur (commented):
LGTM. I can reproduce the failure locally. Looks like `mino/minogrpc/controller/cert.key` needs to be removed sometimes because tests leave it laying around.

lourkeur (approved):

internal/testing/fake/store.go
```diff
@@ -130,7 +130,13 @@ func (tx dbTx) GetBucket(name []byte) kv.Bucket {
...
 // GetBucketOrCreate implements kv.WritableTx.
 func (tx dbTx) GetBucketOrCreate(name []byte) (kv.Bucket, error) {
-	return tx.GetBucket(name), tx.err
+	bucket, found := tx.buckets[string(name)]
+	if !found {
+		bucket = NewBucket()
+		tx.buckets[string(name)] = bucket
+	}
+
+	return bucket, tx.err
```
jbsv, line 139: I wonder if "tx.err" will ever contain an error, as we handle the error case through "found" ? And if it never contains an error, it would be less misleading to return "nil".

mino/minogrpc/controller/mod_test.go
```diff
@@ -418,6 +422,31 @@ func (badReader) Read([]byte) (int, error) {
...
+		N:    big.NewInt(65536),
+	}
+}
+
+func (b badCurve) IsOnCurve(x, y *big.Int) bool {
+	return false
+}
+
+func (b badCurve) Add(x1, y1, x2, y2 *big.Int) (x, y *big.Int) {
+	panic("unsupported")
```
jbsv, line 439: would being more verbose help here ? Like "bad curve is intentionally unsupported".

jbsv (commented):
Besides a couple of questions, LGTM

mino/minogrpc/controller/mod_test.go
pierluca: In this case it's more of a "not implemented" , because I didn't need to for my test , but point taken 😅

PR #253, opened 2023-05-21
add bls PublicKey GetPoint

lourkeur:
Simply an accessor. This would help with the DKG for F3B-IBE.

### DIFF
--- a/crypto/bls/bls_test.go
+++ b/crypto/bls/bls_test.go
@@ -88,6 +88,13 @@ func TestPublicKey_MarshalText(t *testing.T) {
 	require.EqualError(t, err, fake.Err("couldn't marshal"))
 }
 
+func TestPublicKey_GetPoint(t *testing.T) {
+	point := suite.Point()
+	pk := NewPublicKeyFromPoint(point)
+
+	require.True(t, point.Equal(pk.GetPoint()))
+}
+
 func TestPublicKey_String(t *testing.T) {
 	signer := Generate()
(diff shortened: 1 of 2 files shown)


pierluca:
This seems to break the PublicKey abstraction, which arguably is only useful to hide underlying the kyber.Point.
Can you clarify why you need this, the problem you're trying to solve, etc. ?

crypto/bls/bls_test.go
```diff
@@ -88,6 +88,13 @@ func TestPublicKey_MarshalText(t *testing.T) {
 	require.EqualError(t, err, fake.Err("couldn't marshal"))
 }
 
+func TestPublicKey_GetPoint(t *testing.T) {
+	point := suite.Point()
+	pk := PublicKey{point: point}
```
jbsv, line 93:
```suggestion
	pk := NewPublicKeyFromPoint(point)
```

jbsv: Should be the same test as [for the ed25519 implementation](https://github.com/dedis/dela/blob/6cd3d62903806f7863ff29a75c12c6c6bfe06085/crypto/ed25519/ed25519_test.go#L38), in which case the GetPoint() is needed.

jbsv (changes requested):
align test with ed25519

pierluca:
Thank you @lourkeur for the clarification.
That makes sense 👍 
This works for me, the failing test is unrelated to this PR.

jbsv (approved):

PR #255, opened 2023-06-13
Reencrypt cli

jbsv:
- cleaned up the Actor interface to just keep the reencrypt.
- added a shell interface to be able to reencrypt a message through the CLI.

### DIFF
--- a/dkg/dkg.go
+++ b/dkg/dkg.go
@@ -26,10 +26,15 @@ type Actor interface {
 
 	// Encrypt encrypts the given message into kyber points
 	// using the DKG public key
-	Encrypt(message []byte) (K, C kyber.Point, remainder []byte, err error)
+	Encrypt(message []byte) (K kyber.Point, Cs []kyber.Point, err error)
+
 	// Decrypt decrypts a pair of kyber points into the original message
 	// using the DKG internal private key
-	Decrypt(K, C kyber.Point) ([]byte, error)
+	Decrypt(K kyber.Point, Cs []kyber.Point) ([]byte, error)
+
+	// Reencrypt reencrypts generate a temporary key from the public key
+	// to be able to decrypt the message by the user's private key
+	Reencrypt(K kyber.Point, PK kyber.Point) (XhatEnc kyber.Point, err error)
 
 	// Reshare recreates the DKG with an updated list of participants.
 	Reshare(co crypto.CollectiveAuthority, newThreshold int) error
--- a/dkg/dkg.go
+++ b/dkg/dkg.go
@@ -27,7 +27,8 @@ type Actor interface {
 	// Encrypt encrypts the given message into kyber points
 	// using the DKG public key
 	// where K is the ephemeral DH (Diffie-Hellman) public key
-	// and Cs is the resulting, encrypted message
+	// and Cs is the resulting, encrypted points
+	// or an error if encryption is not possible.
 	Encrypt(message []byte) (K kyber.Point, Cs []kyber.Point, err error)
 
 	// Decrypt decrypts a ciphertext (composed of a K and an array of C's)
--- a/dkg/dkg.go
+++ b/dkg/dkg.go
@@ -46,15 +46,4 @@ type Actor interface {
 	// VerifiableDecrypt decrypts a pair of kyber points into the original message
 	// using the DKG internal private key and a proof of work algorithm.
 	VerifiableDecrypt(ciphertexts []types.Ciphertext) ([][]byte, error)
-
-	// EncryptSecret encrypts the given message the calypso way
-	EncryptSecret(message []byte) (U kyber.Point, Cs []kyber.Point)
-
-	// ReencryptSecret is the first version of the calypso
-	// reencryption algorithm in DELA
-	ReencryptSecret(U kyber.Point, Pk kyber.Point) (XhatEnc kyber.Point, err error)
-
-	// DecryptSecret is a helper function to decrypt a secret message previously
-	// encrypted and reencrypted by the DKG
-	DecryptSecret(Cs []kyber.Point, XhatEnc kyber.Point, Sk kyber.Scalar) (message []byte, err error)
 }
--- a/dkg/dkg.go
+++ b/dkg/dkg.go
@@ -34,7 +34,7 @@ type Actor interface {
 	// Decrypt decrypts a ciphertext (composed of a K and an array of C's)
 	// into the original message using the DKG internal private key
 	// where K is the ephemeral DH (Diffie-Hellman) public key
-	// and Cs is the encrypted message
+	// and Cs is the encrypted points
 	Decrypt(K kyber.Point, Cs []kyber.Point) ([]byte, error)
 
 	// Reencrypt reencrypts generate a temporary key from the public key
--- a/dkg/dkg.go
+++ b/dkg/dkg.go
@@ -26,10 +26,14 @@ type Actor interface {
 
 	// Encrypt encrypts the given message into kyber points
 	// using the DKG public key
+	// where K is the ephemeral DH (Diffie-Hellman) public key
+	// and Cs is the resulting, encrypted message
 	Encrypt(message []byte) (K kyber.Point, Cs []kyber.Point, err error)
 
-	// Decrypt decrypts a pair of kyber points into the original message
-	// using the DKG internal private key
+	// Decrypt decrypts a ciphertext (composed of a K and an array of C's)
+	// into the original message using the DKG internal private key
+	// where K is the ephemeral DH (Diffie-Hellman) public key
+	// and Cs is the encrypted message
 	Decrypt(K kyber.Point, Cs []kyber.Point) ([]byte, error)
 
 	// Reencrypt reencrypts generate a temporary key from the public key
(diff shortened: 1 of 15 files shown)


dkg/pedersen/dkg.go
```diff
@@ -282,6 +282,7 @@ func (s *instance) deal(ctx context.Context, out mino.Sender) error {
 	}
 
 	participants := s.startRes.getParticipants()
+	//	for _, to := range participants {
```
pierluca, line 285: commented out code

dkg/pedersen/json/json_test.go
```diff
@@ -312,7 +357,65 @@ func TestMessageFormat_Decode(t *testing.T) {
...
+	data = []byte(fmt.Sprintf(`{"ReencryptReply":{"Pubk":"%s","UiI":13,"UiV":"%s","Ei":[],"Fi":"%s"}}`,
+		testPoint, testPoint, testPoint))
+	_, err = format.Decode(ctx, data)
+	require.ErrorContains(t, err, "couldn't unmarshal Ei")
+
+	data = []byte(fmt.Sprintf(`{"ReencryptReply":{"Pubk":"%s","UiI":13,"UiV":"%s","Ei":"%s","Fi":[]}}`,
+		testPoint, testPoint, testPoint))
+	_, err = format.Decode(ctx, data)
+	require.ErrorContains(t, err, "couldn't unmarshal Fi")
+
```
pierluca, line 415: spurious newline

dkg/pedersen/pedersen.go
```diff
@@ -195,32 +195,53 @@ func (a *Actor) GetPublicKey() (kyber.Point, error) {
...
+
+	// S: ephemeral DH shared secret
+	S := suite.Point().Mul(r, pubK)
+	dela.Logger.Debug().Msgf("S: %v", S.String())
+
+	for len(msg) > 0 {
+		kp := suite.Point().Embed(msg, suite.RandomStream())
+		dela.Logger.Debug().Msgf("kp: %v", kp.String())
+
+		// message blinded with secret
```
pierluca, line 229: shouldn't this be part of the documentation detailing the return values ?

dkg/pedersen/pedersen.go
```diff
@@ -195,32 +195,53 @@ func (a *Actor) GetPublicKey() (kyber.Point, error) {
...
+
 	// ElGamal-encrypt the point to produce ciphertext (K,C).
-	k := suite.Scalar().Pick(random.New())             // ephemeral private key
-	K = suite.Point().Mul(k, nil)                      // ephemeral DH public key
-	S := suite.Point().Mul(k, a.startRes.getDistKey()) // ephemeral DH shared secret
-	C = S.Add(S, M)                                    // message blinded with secret
+	// r: ephemeral, random private key
+	r := suite.Scalar().Pick(suite.RandomStream())
+	// K: ephemeral DH public key
+	// note K = U in cothority
```
pierluca, line 214: shouldn't this be part of the documentation detailing the return values ?

dkg/pedersen/types/messages.go
```diff
@@ -447,16 +447,16 @@ func (req DecryptRequest) Serialize(ctx serde.Context) ([]byte, error) {
 // ReencryptRequest share sent to a node in order to
 // reencrypt a secret with the given public key
 type ReencryptRequest struct {
-	// U is the point from the write-request
-	U kyber.Point
-	// Xc is the public key of the reader
+	// K is the point from the writer
```
pierluca, line 450: `writer` might be unclear here.
Is this point the ciphertext or the randomness ?

dkg/pedersen/resharing_test.go
```diff
@@ -79,9 +79,8 @@ func TestResharing_minoch(t *testing.T) {
 	// Encrypt a message with the old committee public key. The new committee
 	// should be able to decrypt it successfully
 	message := []byte(testMessage)
-	K, C, remainder, err := actorsOld[0].Encrypt(message)
+	K, C, err := actorsOld[0].Encrypt(message)
```
pierluca, line 82: Here and below the `C` are supposed to be `Cs` , no?

dkg/pedersen/reencrypt.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 type onChainSecret struct {
-	U    kyber.Point // U is the random part of the encrypted secret
+	K    kyber.Point // K is the random part of the encrypted secret
```
pierluca, line 16: K is typically used for keys. Why not keep `U` (or something else) for randomness ?

dkg/dkg.go
```diff
@@ -26,30 +26,28 @@ type Actor interface {
 
 	// Encrypt encrypts the given message into kyber points
 	// using the DKG public key
-	Encrypt(message []byte) (K, C kyber.Point, remainder []byte, err error)
-	// Decrypt decrypts a pair of kyber points into the original message
-	// using the DKG internal private key
-	Decrypt(K, C kyber.Point) ([]byte, error)
+	// where K is the ephemeral DH (Diffie-Hellman) public key
```
pierluca, line 29: I'm a bit confused.
Is this `K` the same as the renamed `U`, that referred to randomness ? 
Or is randomness here supposed to be the DH public key ?

dkg/dkg.go
pierluca, line 1: Despite my confusion about the variables, this documentation looks much better 👍

dkg/dkg.go
```diff
@@ -26,30 +26,28 @@ type Actor interface {
...
-	// VerifiableEncrypt encrypts the given message into kyber points
-	// using the DKG public key and proof of work algorithm.
+	// VerifiableEncrypt encrypts the given message into a ciphertext
+	// using the DKG public key and a verifiable encryption function.
 	VerifiableEncrypt(message []byte, GBar kyber.Point) (ciphertext types.Ciphertext, remainder []byte, err error)
 
 	// VerifiableDecrypt decrypts a pair of kyber points into the original message
-	// using the DKG internal private key and a proof of work algorithm.
+	// using the DKG internal private key and a verifiable encryption function.
 	VerifiableDecrypt(ciphertexts []types.Ciphertext) ([][]byte, error)
```
pierluca, line 52: This verifiable encrypt/decrypt now is incompatible with `Encrypt` and `Decrypt` though, isn't it ? 
Isn't this a bit problematic, in terms of the state we're leaving Dela in ? 🤔

dkg/pedersen/controller/action.go
```diff
@@ -230,57 +230,131 @@ func (a decryptAction) Execute(ctx node.Context) error {
...
+}
+
+func decodeEncrypted(str string) (k kyber.Point, cs []kyber.Point, err error) {
+	parts := strings.Split(str, separator)
+	if len(parts) < 2 {
+		return nil, nil, xerrors.Errorf("malformed encoded: %s", str)
+	}
+
+	for i, p := range parts {
+		if i == 0 {
```
pierluca, line 325: This `if/else` looks horrible 😅  Can't we just process i==0 / `K` before iterating and keep the loop cleaner and focused only on the Cs ?

dkg/pedersen/controller/action.go
```diff
@@ -230,57 +230,131 @@ func (a decryptAction) Execute(ctx node.Context) error {
...
+		if err != nil {
+			return "", xerrors.Errorf("failed to marshal c: %v", err)
+		}
+		encoded += separator + hex.EncodeToString(cbuff)
+	}
+
+	return encoded, nil
+}
+
+func decodeEncrypted(str string) (k kyber.Point, cs []kyber.Point, err error) {
```
pierluca, line 318: the function is quite long, I'd avoid using named return variables, declaring them properly and preallocating the right size for `cs`.

dkg/pedersen/pedersen_test.go
```diff
@@ -470,3 +479,44 @@ type fakeSigner struct {
 func (s fakeSigner) GetPublicKey() crypto.PublicKey {
 	return ed25519.NewPublicKeyFromPoint(s.pubkey)
 }
+
+// decryptReencrypted helps to decrypt a reencrypted message.
+func decryptReencrypted(Cs []kyber.Point, XhatEnc kyber.Point, dkgPk kyber.Point, Sk kyber.Scalar) (msg []byte, err error) {
```
pierluca, line 484: Could we spin this out in its own library ?

dkg/pedersen/json/json.go
```diff
@@ -105,13 +105,13 @@ type VerifiableDecryptReply struct {
 }
 
 type ReencryptRequest struct {
-	U    []byte
+	K    []byte
```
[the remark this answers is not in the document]
jbsv: they'are the same. U is the cothority version of K (K as in types.Ciphertext)

dkg/dkg.go
jbsv: I see VerifiableEncrypt / VerifiableDecrypt working in pairs. Same for Encrypt / Decrypt / Reencrypt. You cannot use VerifiableEncrypt with Decrypt anyway.

dkg/pedersen/controller/action.go
jbsv: Not sure if it is less error prone to use 'parts[1:]' for the range ?

dkg/pedersen/pedersen_test.go
jbsv: I'll do that in next step.

dkg/dkg.go
jbsv: As far as I understand it, it's all the same. This comment comes from cothorithy. I thought it would be good to have this info here.

PR #257, opened 2023-07-18
Reencrypt cli2

jbsv:
fixed the threshold issue for re-encryption

### DIFF
--- a/dkg/pedersen/controller/action.go
+++ b/dkg/pedersen/controller/action.go
@@ -44,6 +44,9 @@ func (a setupAction) Execute(ctx node.Context) error {
 	}
 
 	t := ctx.Flags.Int("threshold")
+	if t < 2 || t > co.Len() {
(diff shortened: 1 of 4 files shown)


dkg/pedersen/controller/action.go
```diff
@@ -44,6 +44,9 @@ func (a setupAction) Execute(ctx node.Context) error {
 	}
 
 	t := ctx.Flags.Int("threshold")
+	if t < 2 || t > co.Len() {
+		return xerrors.Errorf("thresold (%d) needs to be between 2 and %d", t, co.Len())
+	}
```
pierluca, line 49: You've convinced me this can be removed to avoid duplicating logic.

jbsv: actor.Setup() will return an error if threshold is 0 (i.e. not in the boundary), which will report an error when executing the command

Issue #258, opened 2023-07-24
Value contract allows overwriting arbitrary keys in store


pierluca:
The value contract does not prefix the keys provided by the smart contract user to store a value.
As such, a malicious user could overwrite any key/value combination in the store, including other contracts'.

This is a non-critical because the value contract is a toy contract and not used in production, but it should be addressed nonetheless.

PR #260, opened 2023-08-04
Added prefix for value contract storage

pierluca:
There's a weakness in Dela (#258) whereby the (existing) smart contracts store key=values without any prefix for the keys in a shared key/value store. Effectively this would allow any malicious user to overwrite other smart contracts' data.

This fixes it for the `value` contract.

We probably need to do something similar for https://github.com/dedis/d-voting

@PascalinDe @lanterno for your information, the current key/value store is limited to keys of 32 bytes. Do reach out if adding a prefix could be a problem for D-Voting (we haven't checked).

### DIFF
--- a/contracts/value/value.go
+++ b/contracts/value/value.go
@@ -44,6 +44,9 @@ const (
 	// credentialAllCommand defines the credential command that is allowed to
 	// perform all commands.
 	credentialAllCommand = "all"
+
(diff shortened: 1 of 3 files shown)


PascalinDe (approved):
LGTM

pierluca:
I haven't checked the CI, but I expect some integration tests will be broken.
More work is needed, but it should give a pointer to the general direction.

PR #261, opened 2023-08-04
Extend crypto to support SHA3-224

pierluca:
Sorry, this also contains some re-formatting :-/


crypto/hash_test.go
```diff
@@ -7,6 +7,18 @@ import (
 )
```
PascalinDe, line 8: shouldn't we keep a test for the old deprecated version as well as long as it's supposed to work?

PascalinDe (changes requested):

jbsv (approved):

crypto/hash_test.go
pierluca: Good point, thanks !

PascalinDe (commented):
LGTM now :+1:

PR #263, opened 2023-08-08
More realistic fake.InMemorySnapshot

pierluca:
Some tests were becoming hard to write with the super-simple fake `InMemorySnapshot`.
This should make things easier, but carries the risk of breaking things.
This draft PR was created to let the CI scream at us while we proceed with other developments.


jbsv (approved):
LGTM

PR #272, opened 2023-09-26
Add working --serveTLS

ineiti:
Instead of using `--noTLS` to indicate not to use TLS, changed the flag to `--serveTLS`. Now:

- if a certificate and a chain is given, `--serveTLS` is enabled
- if `--serveTLS` is enabled w/o a certificate, a self-signed certificate is created
- if the remote point starts with `https`, a TLS connection is established, and the certificate is verified using public CAs used in the system


mino/minogrpc/server.go
```diff
@@ -11,13 +11,12 @@ import (
 	"context"
 	"crypto"
 	"crypto/ecdsa"
-	"crypto/rand"
 	"crypto/tls"
 	"crypto/x509"
+	"google.golang.org/grpc/credentials/insecure"
+	"strings"
```
pierluca, line 17: we're using goimports formatting, which isn't respected here.

mino/minogrpc/server.go
```diff
@@ -702,16 +678,23 @@ func (mgr *connManager) Acquire(to mino.Address) (grpc.ClientConnInterface, erro
...
+	if mgr.serveTLS {
 		ta, err := mgr.getTransportCredential(to)
 		if err != nil {
 			return nil, xerrors.Errorf("failed to retrieve transport credential: %v", err)
 		}
 
 		opts = append(opts, grpc.WithTransportCredentials(ta))
+	} else {
+		// If the remote is accessible via port 443, we suppose it is TLS terminated and signed by a
+		// CA available on the system.
```
pierluca, line 692: this assumption might easily break in our own deployments.
that was the whole point of passing the certificate hash along with the token when joining nodes 😕

mino/minogrpc/server.go
```diff
@@ -535,19 +538,41 @@ func (o *overlay) GetCertificateStore() certs.Storage {
...
+		switch addr.Scheme {
+		case "http":
+			host += ":80"
+		case "https":
+			host += ":443"
+		default:
+			return xerrors.Errorf("address doesn't contain a port and uses "+
+				"an unknown scheme: %s", addr.Scheme)
+		}
+	}
```
pierluca, line 560: why is this needed ? Effectively the code is adding default ports for those schemes, that's normally not needed 🤔

mino/minogrpc/server.go
```diff
@@ -476,42 +499,22 @@ func newOverlay(tmpl *minoTemplate) (*overlay, error) {
...
-		if err != nil {
-			return nil, xerrors.Errorf("while loading cert: %v", err)
-		}
-
-		if cert == nil {
-			err = o.makeCertificate()
-			if err != nil {
-				return nil, xerrors.Errorf("certificate failed: %v", err)
-			}
-		}
```
pierluca, line 508: good call on moving all this stuff ahead of the overlay creation, it feels cleaner !

pierluca (commented):
I have some serious doubts about these changes.

I might have missed something (it's late and it was a long day), but if I understand correctly, we're introducing (or revamping) insecure defaults, and the solution to handling reverse proxies is removing a MitM protection, relying entirely on certificate authorities to verify remote node certificates.

If my understanding is correct, this would (at the very least) break a few of our own internal (experimental) deployments, which aren't always exposed to the internet. It would also make it harder to catch configuration mistakes.

I'm happy to have a quick (offline) chat about this if it helps move the functionality along.

mino/minogrpc/server.go
ineiti: I'm never sure what Goland does, sorry. I'll try to convince it to use `goimports`. Do you have a common formatter you're using?

mino/minogrpc/mod.go
```diff
@@ -118,16 +122,63 @@ type Minogrpc struct {
...
+	certs    certs.Storage
+	secret   interface{}
+	public   interface{}
+	curve    elliptic.Curve
+	random   io.Reader
+	cert     *tls.Certificate
+	serveTLS bool
+}
+
+func (m *minoTemplate) makeCertificate() error {
```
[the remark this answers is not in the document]
ineiti: I only moved this from the `overlay` to `minoTemplate` to simplify `newOverlay`.

mino/minogrpc/mod.go
```diff
@@ -223,28 +274,28 @@ func NewMinogrpc(listen net.Addr, public *url.URL, router router.Router, opts ..
 		grpc.StreamInterceptor(otgrpc.OpenTracingStreamServerInterceptor(tracer, otgrpc.SpanDecorator(decorateServerTrace))),
 	}
 
-	if !tmpl.useTLS {
-		dela.Logger.Warn().Msg("⚠️ running in insecure mode, you should not " +
-			"publicly expose the node's socket")
+	if !tmpl.serveTLS && public.Scheme != "https" {
+		dela.Logger.Warn().Msg("⚠️ running in insecure mode and no TLS endpoint, you should not " +
+			"publicly expose the node's socket without TLS")
```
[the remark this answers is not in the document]
ineiti: remove check for `public.Scheme` and reword warning for students

ineiti:
After discussion:

By default:
- `makeCertificate` and listen on TLS

Or one of:
- give `--certKey`
- give both `--certKey` and `--certChain`
- `--noTLS` to remove listening on TLS

ineiti:
Done in #280 and #282

PR #275, opened 2023-11-21
Minor changes

ineiti:
This PR includes mostly changed comments and some corrections to the documentation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,9 +22,26 @@ vet: tidy
 	go install ./internal/mcheck && \
 	go vet -vettool=`go env GOPATH`/bin/mcheck -commentLen -ifInit ./...
 
+tests:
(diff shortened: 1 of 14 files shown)


ineiti:
Hmm - because I don't have the `SONAR_TOKEN` variable, it cannot send the results to the sonarcloud...

Makefile
pierluca, line 1: Is this really simpler than addressing the underlying causes of the test flakiness ? :-/ 
I tend to get a little worried when I start seeing things like this in the Makefile 😅

ineiti: This work is left as an exercise to the reader...

Yes, it would be better, but I did spend quite some time on getting the tests to run, and after some time I didn't advance, so I came up with this, which is in fact a copy of what I did for some tests in the cothority :)

ineiti:
Closing this for the new PRs

PR #283, opened 2024-04-03
optimize chain

PascalinDe:
* add fastsync protocol
* fix comparisons/updates of block indices and nonces


PascalinDe:
@ineiti I wasn't sure how to split this up in smaller PRs as they seem to more or less revolve around the same issue (syncing) with maybe one or two commits that address some other issue, I let you split this up more sensibly if need be

.github/workflows/go_lint.yml
```diff
@@ -2,7 +2,7 @@ name: Go lint
 
 on:
   push:
-    branches: [ master ]
+    branches: [ main ]
```
ineiti, line 5: @pierluca @jbsv - not sure if you want to switch to `main` branch instead of `master`. We switched all our repos to `main`: https://www.theserverside.com/feature/Why-GitHub-renamed-its-master-branch-to-main

core/execution/native/native.go
```diff
@@ -84,6 +84,8 @@ func (ns *Service) Execute(snap store.Snapshot, step execution.Step) (execution.
 	err := contract.Execute(snap, step)
 	if err != nil {
 		res.Accepted = false
+		// LG: DEBUG - I'd like to keep this commented line, as it helps debugging.
+		// res.Message = fmt.Sprintf("%+v", err)
```
ineiti, line 88: @ineiti TODO: need to remove these lines...

ineiti (changes requested):
Just one comment that needs to be removed - @PascalinDe do you want to patch the commit? Or shall I do it?

ineiti:
Closed in favor of #285

PR #297, opened 2025-02-07
Updating grpc

ineiti:
- updating grpc packages to latest version so that grpc compiles correctly
- adding proto-files in minogrpc/ptypes
- fixing not exported `stop` method in `minows`

The `cli/node/memcoin` test still fails.


jbsv:
closing this pull request as the fixed of the protobuf generation as been added to dela

PR #300, opened 2025-02-08
Fix lint warnings

jbsv:
after setting up the golangci tool, attemp to reduce lint warnings


ineiti:
Where are the LLMs when they could be useful?

PR #303, opened 2025-04-07
Upgrade libp2p

jbsv:


.golangci.yml
```diff
@@ -97,11 +97,11 @@ linters:
     disable-all: true
     enable:
         ## enabled by default
-        - errcheck # checking for unchecked errors, these unchecked errors can be critical bugs in some cases
+        #- errcheck # checking for unchecked errors, these unchecked errors can be critical bugs in some cases
         - gosimple # specializes in simplifying a code
-        - govet # reports suspicious constructs, such as Printf calls whose arguments do not align with the format string
+        #- govet # reports suspicious constructs, such as Printf calls whose arguments do not align with the format string
```
pierluca, line 102: I would argue that `govet` and `errcheck` are important to avoid rather fundamental bugs, do they still report issues?

Makefile
```diff
@@ -1,20 +1,21 @@
...
+all: lint test
 	@echo "Done with the standard checks"
 
 tidy:
-	go mod tidy
+	@go mod tidy -go="1.23.8"
 
 generate: tidy
-	go get -u google.golang.org/protobuf/cmd/protoc-gen-go@v1.36.5
+	go get -u google.golang.org/protobuf/cmd/protoc-gen-go@v1.5.1
```
pierluca: I'm wondering if we really need to revert to such an old version?
protoc-gen-go and protoc-gen-go-grpc versions are not aligned, this seems likely unnecessary and potentially problematic 🤔
protoc-gen-go-grpc 1.5.1 was released >=2024.
protoc-gen-go 1.5.1 was released <= 2020

cli/node/memcoin/mod_test.go
```diff
@@ -52,7 +53,7 @@ func TestMemcoin_Scenario_SetupAndTransactions(t *testing.T) {
 		wg.Wait()
 	}()
 
-	require.True(t, waitDaemon(t, []string{node1, node2, node3}), "daemon failed to start")
+	require.True(t, waitDaemon([]string{node1, node2, node3}), "daemon failed to start")
```
pierluca, line 56: in a future PR, we might want to change the API such that the call looks like
```suggestion
	requireDaemonStart(t, node1, node2, node3)
```
given `t` and `"daemon failed to start"` are the same in all tests.

core/ordering/cosipbft/blockstore/genesis_test.go
```diff
@@ -88,17 +83,19 @@ func TestGenesisDiskStore_Load(t *testing.T) {
 	store.fac = fake.MessageFactory{}
 	err = store.Load()
 	require.EqualError(t, err, "unsupported message 'fake.Message'")
-}
 
-func TestGenesisDiskStore_Set(t *testing.T) {
-	dir, err := os.MkdirTemp(os.TempDir(), "dela-blockstore-")
+	err = db.Close()
 	require.NoError(t, err)
```
pierluca: This is inconsistent with the fix below, that looks better.
Do you want to `db.Close()` every time ? If so the `defer db.Close()` would make sense and avoid the DB remaining open if one of the other assertions fail.

core/ordering/cosipbft/pbft/pbft_test.go
```diff
@@ -897,7 +895,7 @@ func makeTree(t *testing.T) (hashtree.Tree, kv.DB, func()) {
 	stage, err := tree.Stage(func(store.Snapshot) error { return nil })
 	require.NoError(t, err)
 
-	return stage, db, func() { os.RemoveAll(dir) }
+	return stage, db
```
pierluca, line 898: I don't see why we couldn't do the `db.Close()` in the clean function here, instead of changing it everywhere - it likely would've been more maintainable in case further clean-ups were needed 😅 
(like you did in pow_test.go)

core/store/hashtree/binprefix/binprefix_test.go
```diff
@@ -317,14 +316,13 @@ func TestWritableMerkleTree_Delete(t *testing.T) {
...
-func makeDB(t *testing.T) (kv.DB, func()) {
-	dir, err := os.MkdirTemp(os.TempDir(), "dela-pow")
-	require.NoError(t, err)
+func makeDB(t *testing.T) kv.DB {
+	dir := t.TempDir()
 
 	db, err := kv.New(filepath.Join(dir, "test.db"))
 	require.NoError(t, err)
 
-	return db, func() { os.RemoveAll(dir) }
```
pierluca, line 327: same comment, the pow_test.go approach seems cleaner.

dkg/pedersen/json/json.go
```diff
@@ -476,8 +478,7 @@ func encodeReshare(msg types.Reshare) (Message, error) {
 	return Message{Reshare: &dr}, nil
 }
 
-func (f msgFormat) decodeReshare(ctx serde.Context,
-	msg *Reshare) (serde.Message, error) {
+func (f msgFormat) decodeReshare(msg *Reshare) (serde.Message, error) {
```
pierluca, line 481: why remove the context here and replace it with an ignored parameter (`_`) everywhere else? 
It seems inconsistent 🤔

pierluca (changes requested):
Mostly LGTM, but there are a few obvious things that need fixing :)

PR #306, opened 2025-06-03
Minows uses a DB to handle keys

jbsv:


test/cosidela_test.go
```diff
@@ -116,7 +119,7 @@ func newDelaNode(t require.TestingT, path string, port int, kind string) dela {
 		listen, err := ma.NewMultiaddr(fmt.Sprintf("/ip4/127.0.0.1/tcp/%d/ws", port))
 		require.NoError(t, err)
 
-		onet, err = minows.NewMinows(listen, nil, &db)
+		onet, err = minows.NewMinows(listen, nil, db, port)
```
ineiti: Why is the port used as index for the key?

ineiti:
Why? What is the reason for this PR? Is there an issue attached to it?

From what I see, the private key index is only used in the tests. I propose to change the tests to use one db per instance, so you don't need to have a key-index. Because you'll also have other data in the db, so you'll need multiple dbs anyway.

